PR #1, opened 2020-09-20
now reads from config file

alx365:

PR #22, opened 2020-10-18
Merge pull request #1 from elkowar/master

alx365:
a


elkowar:
You should probably fix ur git workflow, mate :100:

alx365:
that was also the wrong base :/

PR #31, opened 2020-10-19
added a always-there variable called window_name, that contains.. who…

alx365:
… would have guessed.. the window name - LegendOfMiracles


alx365:
closed? what does that mean?

PR #41, opened 2020-10-24
Combo box and as_vec()

alx365:


src/value/primitive.rs
```diff
@@ -91,4 +92,27 @@ impl PrimitiveValue {
...
+                    let it = items.remove((times_ran + 1) - removed);
+                    items[times_ran - removed] += &it;
+                    removed += 1;
+                }
+            }
+            Ok(items)
+        }
+        None => Err(anyhow!("Is your array built like this: '[these,are,items]'?")),
+    }
 }
```
elkowar, line 118: I added some unit tests, a few of which fail. You should add these and also fix the code :D 
```suggestion
}

#[cfg(test)]
mod test {
    use super::*;
    use pretty_assertions::assert_eq;
    #[test]
    fn test_parse_vec() {
        assert_eq!(
            Vec::<String>::new(),
            parse_vec("[]".to_string()).unwrap(),
            "should be able to parse empty lists"
        );
        assert_eq!(
            vec!["hi"],
            parse_vec("[hi]".to_string()).unwrap(),
            "should be able to parse single element list"
        );
        assert_eq!(
            vec!["hi", "ho", "hu"],
            parse_vec("[hi,ho,hu]".to_string()).unwrap(),
            "should be able to parse three element list"
        );
        assert_eq!(
            vec!["hi,ho"],
            parse_vec("[hi\\,ho]".to_string()).unwrap(),
            "should be able to parse list with escaped comma"
        );
        assert_eq!(
            vec!["hi,ho", "hu"],
            parse_vec("[hi\\,ho,hu]".to_string()).unwrap(),
            "should be able to parse two element list with escaped comma"
        );
        assert!(parse_vec("".to_string()).is_err(), "Should fail when parsing empty string");
        assert!(
            parse_vec("[a,b".to_string()).is_err(),
            "Should fail when parsing unclosed list"
        );
        assert!(parse_vec("a]".to_string()).is_err(), "Should fail when parsing unopened list");
    }
}

```

src/widgets/widget_definitions.rs
```diff
@@ -146,6 +147,29 @@ pub(super) fn resolve_orientable_attrs(bargs: &mut BuilderArgs, gtk_widget: &gtk
...
+fn build_gtk_combo_box(bargs: &mut BuilderArgs) -> Result<gtk::ComboBoxText> {
+    let gtk_widget = gtk::ComboBoxText::new();
+    let on_change_handler_id: Rc<RefCell<Option<glib::SignalHandlerId>>> = Rc::new(RefCell::new(None));
+    resolve_block!(bargs, gtk_widget, {
+        // @prop items
+        prop(items: as_vec) {
+            for i in items {
+            gtk_widget.append_text(&i);
+            }
+        },
```
elkowar, line 160: This will not correctly handle `items` being changed. currently, it would just add elements to the end, rather than replacing the contents. you'll need to reset before appending, first.

src/widgets/widget_definitions.rs
```diff
@@ -146,6 +147,29 @@ pub(super) fn resolve_orientable_attrs(bargs: &mut BuilderArgs, gtk_widget: &gtk
...
+        // @prop onchange - runs the code when a item was selected, replacing {} with the item as a string
+        prop(onchange: as_string) {
+            let old_id = on_change_handler_id.replace(Some(
+                gtk_widget.connect_changed(move |gtk_widget| {
+                    run_command(&onchange, gtk_widget.get_active_text().unwrap());
```
elkowar, line 165: This `unwrap` will panic if no value is selected. I'd guess that that's a realistic state to be in, thus you should _definitely_ not unwrap here

src/widgets/widget_definitions.rs
alx365: fixed

src/widgets/widget_definitions.rs
alx365: I tested it, and the even only gets triggered, when something is selected.. that means if it has a value, thus the unwrap should *definitely* be ok

alx365:
epic i made a typo in commit message. pls ignore and merge

src/widgets/widget_definitions.rs
elkowar: Still not a fan, why not just do `unwrap_or_default` to have an empty string in case it _does_ happen?

alx365: understandable

alx365: fixed with `"".into` in `.unwrap_or()`

elkowar:
Looks good now! Thanks for the contribution!

Issue #12, opened 2020-10-14
Support Multi-file configuration


elkowar:
Eww should support an `<includes>`  block that allows users to include XML from other files.

This block should be placed at the same level as `<definitions>`, `<variables>` and `<windows>`, and have the following shape:
```xml
<includes>
  <file path="./something.xml"/>
  <file path="./somethingelse.xml"/>
</includes>
```

every one of these files should follow the same structure as the main `eww.xml` file, and may themselves contain other includes. 

At parse time, all the includes should first be parsed into their own `EwwConfig` instance, which should then be merged into one combined `EwwConfig`.

For error messages to stay helpful, the text position of each `WidgetUse` should be expanded into a new struct which also stores the path to the file it was read from. This should then be included in the `Display` implementation of said struct. 

In a later version of this, a namespacing system should be introduced, but that is not part of this issue.

**Acceptance Criteria**
- The EwwConfig parser is expanded to allow for a new block: `<includes>`, which contains a list of files to include.
- Included files are parsed as EwwConfigs and then merged together
- Error messages that show line-numbers correctly contain the path to the file the error originated in
- File paths are correctly resolved relative to the current file
- Tests are written to show that the implementation of merging `EwwConfig`s is correct

elkowar:
@druskus20, this may be something for you. Should not be too hard to implement. If you want, we could go over the implementation together ;D

druskus20:
Sure, I can do this. Will let you know if I need some help.

elkowar:
perfect, i'll assign this issue to you ;D

PR #45, opened 2020-10-25
Support Multi-file configuration 

druskus20:
## Description
This PR implements multi-file configuration according to #12.

### Usage

This pr supports an <includes> block that allows users to include XML from other files.

This block should be placed at the same level as <definitions>, <variables> and <windows>, and have the following shape:
```
<includes>
  <file path="./something.xml"/>
  <file path="./somethingelse.xml"/>
</includes>
```

### Showcase
...

## Additional Notes

TODO: 
✓ - The EwwConfig parser is expanded to allow for a new block: <includes>, which contains a list of files to include.
✓ - Included files are parsed as EwwConfigs and then merged together
✓ - Error messages that show line-numbers correctly contain the path to the file the error originated in
✓ - File paths are correctly resolved relative to the current file
✓ - Tests are written to show that the implementation of merging EwwConfigs is correct

## Checklist


src/config/eww_config.rs
```diff
@@ -36,35 +37,29 @@ impl EwwConfig {
...
         let includes =
             match xml.child("includes") {
                 Ok(tag) => tag.child_elements()
                     .map(|child| {
-                        let path = CONFIG_DIR.join(child.attr("path").unwrap());
-                        EwwConfig::read_from_file(path)
+                        let childpath = child.attr("path").unwrap();
+                        let mut basepath = path.as_ref().parent().unwrap();
+                        EwwConfig::read_from_file(basepath.join(childpath))
                     })
```
elkowar: Really not a fan of these unwraps here. you should easily be able to use `?` instead:
```rs
                    .map(|child| {
                        let childpath = child.attr("path")?;
                        let mut basepath = path.as_ref().parent().context("config file had no parent directory")?;
                        EwwConfig::read_from_file(basepath.join(childpath))
                    })
```

src/config/eww_config.rs
```diff
@@ -36,35 +37,29 @@ impl EwwConfig {
...
-            .map(|v| PathBuf::from(v))
-            .unwrap_or_else(|_| PathBuf::from(std::env::var("HOME").unwrap()).join(".config"))
-            .join("eww");
+    pub fn from_xml_element<P: AsRef<std::path::Path>>(xml: XmlElement, path: P) -> Result<Self> {
 
         // !!! This doesnt seem that bad
         let includes =
             match xml.child("includes") {
                 Ok(tag) => tag.child_elements()
                     .map(|child| {
```
elkowar, line 53: Please also verify the tag name here, the same way it is done for other configuration tags. This is to avoid that the user is able to put _any_ tag, no matter it's name in here, as long as it has the correct "path" attribute

src/config/eww_config.rs
```diff
@@ -36,35 +37,29 @@ impl EwwConfig {
...
-                        let path = CONFIG_DIR.join(child.attr("path").unwrap());
-                        EwwConfig::read_from_file(path)
+                        let childpath = child.attr("path").unwrap();
+                        let mut basepath = path.as_ref().parent().unwrap();
+                        EwwConfig::read_from_file(basepath.join(childpath))
                     })
                     .collect::<Result<Vec<_>>>()
                     .context("error parsing include definitions")?,
```
elkowar, line 59: This isn't really only parsing
```suggestion
                    .context("error handling includes")?,
```

elkowar (commented):
I left a couple comments already, nothing major tho. Looking good already!

src/config/eww_config.rs
druskus20: Manually added.

src/config/eww_config.rs
druskus20: Done.

druskus20:
Undid silly stuff

src/config/eww_config.rs
druskus20: Done tho

src/config/eww_config.rs
```diff
@@ -122,3 +149,153 @@ impl EwwConfig {
...
+        };
+
+        let merged_config = EwwConfig::merge_includes(base_config, vec![config1.unwrap(), config2.unwrap()]).unwrap();
+
+        assert_eq!(merged_config.widgets.len(), 2);
+        assert_eq!(merged_config.windows.len(), 2);
+        assert_eq!(merged_config.initial_variables.len(), 2);
+        assert_eq!(merged_config.script_vars.len(), 0);
+    }
+}
```
elkowar, line 301:
```suggestion
```
If i'm not blind this is the same test block as the one above... no need for duplicates (in fact, probably doesn't compile even)

src/config/eww_config.rs
```diff
@@ -50,7 +76,7 @@ impl EwwConfig {
                 Ok((def.name.to_owned(), def))
             })
             .collect::<Result<HashMap<_, _>>>()
-            .context("error parsing window definitions")?;
+            .context(format!("error parsing window definitions: {}", path.display()))?;
```
elkowar, line 79:
```suggestion
            .with_context(|| format!("error parsing window definitions: {}", path.display()))?;
```
with_context makes this more performant, as it only does the string formatting computation when the erorr actually happens.

src/config/eww_config.rs
```diff
@@ -40,7 +66,7 @@ impl EwwConfig {
                 Ok((def.name.clone(), def))
             })
             .collect::<Result<HashMap<_, _>>>()
-            .context("error parsing widget definitions")?;
+            .context(format!("error parsing widget definitions: {}", path.display()))?;
```
elkowar, line 69:
```suggestion
            .with_context(|| format!("error parsing widget definitions: {}", path.display()))?;
```
Same as below

src/config/eww_config.rs
```diff
@@ -23,15 +23,41 @@ pub struct EwwConfig {
...
 
-        let result = EwwConfig::from_xml_element(XmlNode::from(document.root_element()).as_element()?.clone());
+        let result = EwwConfig::from_xml_element(XmlNode::from(document.root_element()).as_element()?.clone(), path.as_ref());
         result
     }
 
-    pub fn from_xml_element(xml: XmlElement) -> Result<Self> {
+    pub fn from_xml_element<P: AsRef<std::path::Path>>(xml: XmlElement, path: P) -> Result<Self> {
+        let path = path.as_ref();
+        let includes = match xml.child("includes").context("Error") {
```
elkowar, line 48:
```suggestion
        let includes = match xml.child("includes") {
```

That's,... unnecessary

src/config/eww_config.rs
```diff
@@ -23,15 +23,41 @@ pub struct EwwConfig {
 }
 
 impl EwwConfig {
+    // TODO: There is definitely a better way to do this with a fold
```
elkowar, line 26:
```suggestion
```
it's fine

elkowar (changes requested):
Couple changes still to do ;D 
Also, a couple doc-comments, specifically on the merge function would be nice.

src/config/eww_config.rs
druskus20: That was definetly an error on my part. Idk how that ended there.  ^^

elkowar (approved):
Thanks, looks good now!

elkowar (approved):
Looks good!

elkowar:
Lol imagine github cli just giving u the wrong number :'(

PR #47, opened 2020-10-26
fixed widget docs issues, with missing descriptions and renamed a widget

alx365:
This only adds stuff to the docs (or the be generated ones :D), i almost had no description at any of the widgets i had added. I also changed a bit of your descriptions. 

This also  renames a widget, to what it actually is, namingly the combo box -> combo box text because that's what the code actually generates.

### DIFF
--- a/src/widgets/widget_definitions.rs
+++ b/src/widgets/widget_definitions.rs
@@ -149,7 +149,7 @@ pub(super) fn resolve_orientable_attrs(bargs: &mut BuilderArgs, gtk_widget: &gtk
 // concrete widgets
 
 /// @widget combo-box-text
-/// @desc A combo box, where one can choose between several items (must be text, combo-box without text is not yet implemented)
+/// @desc A combo box, where one can choose between several items (must be text, combo-box, with other widgets, is not yet implemented)
 fn build_gtk_combo_box_text(bargs: &mut BuilderArgs) -> Result<gtk::ComboBoxText> {
     let gtk_widget = gtk::ComboBoxText::new();
     let on_change_handler_id: Rc<RefCell<Option<glib::SignalHandlerId>>> = Rc::new(RefCell::new(None));
--- a/src/widgets/widget_definitions.rs
+++ b/src/widgets/widget_definitions.rs
@@ -149,7 +149,7 @@ pub(super) fn resolve_orientable_attrs(bargs: &mut BuilderArgs, gtk_widget: &gtk
 // concrete widgets
 
 /// @widget combo-box-text
-/// @desc A combo box, where one can choose between several items (must be text, combo-box, with other widgets, is not yet implemented)
+/// @desc A combo box allowing the user to choose between several items.
 fn build_gtk_combo_box_text(bargs: &mut BuilderArgs) -> Result<gtk::ComboBoxText> {
     let gtk_widget = gtk::ComboBoxText::new();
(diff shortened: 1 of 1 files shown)


src/widgets/widget_definitions.rs
```diff
@@ -247,7 +250,7 @@ fn build_gtk_scale(bargs: &mut BuilderArgs) -> Result<gtk::Scale> {
 }
 
 /// @widget input
-/// @desc an input field. For this to be useful, set `focusable="true"` on the window.
+/// @desc A input field. For this to be useful, set `focusable="true"` on the window.
```
elkowar, line 253:
```suggestion
/// @desc An input field. For this to be useful, set `focusable="true"` on the window.
```

src/widgets/widget_definitions.rs
```diff
@@ -230,7 +233,7 @@ fn build_gtk_color_chooser(bargs: &mut BuilderArgs) -> Result<gtk::ColorChooserW
 }
 
 /// @widget scale extends range
-/// @desc a slider.
+/// @desc A widget with a slider.
```
elkowar, line 236: It is just simply the slider tho :o
```suggestion
/// @desc A slider.
```

src/widgets/widget_definitions.rs
```diff
@@ -319,6 +324,7 @@ fn build_gtk_box(bargs: &mut BuilderArgs) -> Result<gtk::Box> {
 }
 
 /// @widget label
+/// @desc A widget that displays a small to medium amount of text
```
elkowar, line 327: Maybe explain why this is relevant if you can just use inline text too
```suggestion
/// @desc A text widget giving you more control over how the text is displayed
```

src/widgets/widget_definitions.rs
```diff
@@ -289,6 +293,7 @@ fn build_gtk_button(bargs: &mut BuilderArgs) -> Result<gtk::Button> {
 }
 
 /// @widget image
+/// @desc A widget displaying images
```
elkowar, line 296: It only renders a single image 
```suggestion
/// @desc A widget displaying an image
```

src/widgets/widget_definitions.rs
```diff
@@ -148,12 +148,13 @@ pub(super) fn resolve_orientable_attrs(bargs: &mut BuilderArgs, gtk_widget: &gtk
 
 // concrete widgets
 
-/// @widget combo-box
-fn build_gtk_combo_box(bargs: &mut BuilderArgs) -> Result<gtk::ComboBoxText> {
+/// @widget combo-box-text
+/// @desc A combo box, where one can choose between several items (must be text, combo-box, with other widgets, is not yet implemented)
```
elkowar, line 152: I don't think it's necessary to mention "not yet implement" stuff in the docs here, as the API of the combo-box-text is quite obviously text-only

```suggestion
/// @desc A combo box allowing the user to choose between several items.
```

elkowar (changes requested):

src/widgets/widget_definitions.rs
alx365: 🥇

src/widgets/widget_definitions.rs
alx365: i copied that from the gnome docs. pls no bully 
image

src/widgets/widget_definitions.rs
alx365: hm tru

elkowar:
Thanks!

alx365:
**EPIC***

PR #57, opened 2020-11-07
Edit subcommand to edit config files

legendofmiracles:
#  Description
Adds a subcommand to edit the eww.xml and scss file, with checking if the file is correct after it being edited.

## Usage

```
eww-edit 0.1.0
Edit the configuration files

USAGE:
    eww edit [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -e, --editor <editor>    Editor to use
    -f, --file <file>        What file to edit (the main xml file or the main scss file) accepted values are xml or scss
```
Then press x or s depending on what you want to edit.

Change stuff then quit out of the editor.

### Showcase
image
(newer "asking system" to prompt for file in the bottom screenshot)
image

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -42,8 +42,11 @@ pub fn handle_client_only_action(action: ActionClientOnly) -> Result<()> {
                 path = scss_file;
             } else {
                 eprint!("Edit the eww.xml file or the scss file? (X/s) ");
-                let input = input()?;
-                path = if input.to_lowercase() == "s\n" { scss_file } else { xml_file }
+                path = if input()?.to_lowercase() == "s\n" {
+                    scss_file
+                } else {
+                    xml_file
+                }
             }
 
             launch_editor(&editor, path.to_str().unwrap())?;
@@ -51,9 +54,8 @@ pub fn handle_client_only_action(action: ActionClientOnly) -> Result<()> {
                 while let Some(config) = crate::config::EwwConfig::read_from_file(&path).err() {
                     eprintln!("{}", config);
                     eprint!("The config file contains errors, edit again? (Y/n) ");
-                    let input = input()?;
                     // \n is there because input is unsanitized and it still contains the newline
-                    if input.to_lowercase() == "n\n" {
+                    if input()?.to_lowercase() == "n\n" {
                         break;
                     } else {
                         launch_editor(&editor, path.to_str().unwrap())?;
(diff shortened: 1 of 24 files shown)


src/client.rs
```diff
@@ -16,6 +17,66 @@ pub fn handle_client_only_action(action: ActionClientOnly) -> Result<()> {
...
+            } else if file == "scss" {
+                path = scss_file;
+            } else {
+                eprint!("Edit the eww.xml file or the scss file? (X/s) ");
+                path = if input()?.to_lowercase() == "s\n" {
+                    scss_file
+                } else {
+                    xml_file
+                }
+            }
```
elkowar, line 50:
```suggestion
            let (xml_file, scss_file) = config_path()?;
            let path = match file {
              Some("xml") => xml_file,
              Some("scss") => scss_file,
              None => {
                eprint!("Edit the eww.xml file or the scss file? (X/s) ");
                if input()?.to_lowercase() == "s\n" { scss_file }  else { xml_file }
              }
            }
```

src/client.rs
```diff
@@ -16,6 +17,66 @@ pub fn handle_client_only_action(action: ActionClientOnly) -> Result<()> {
...
+
+            launch_editor(&editor, path.to_str().unwrap())?;
+            if path.extension().unwrap() == "xml" {
```
elkowar, line 53: Just outsource all of this into a function and then recursively call the function.... these while loops are really unnecessary.

src/opts.rs
```diff
@@ -32,6 +32,16 @@ pub enum ActionClientOnly {
     /// Print and watch the eww logs
     #[structopt(name = "logs")]
     Logs,
+    /// Edit the configuration files
+    #[structopt(name = "edit")]
+    Edit {
+        /// Editor to use
+        #[structopt(name = "editor", short, long)]
+        editor_arg: Option<String>,
+        /// What file to edit (the main xml file or the main scss file) accepted values are xml or scss
```
elkowar, line 41: A much nicer implementation of this would be to use an enum type with the two options.
As we'll likely be waiting for multi-file config to be finished, using strings will be preferrable if these are done... thus i guess using strings is fine here.

elkowar (changes requested):

elkowar:
This definitely still needs some work. I think we should also generally wait until #45 is done, as that will affect this PR a lot.

src/client.rs
alx365: ok.. tomorrow

src/opts.rs
alx365: wow very amazing suggestion then, if we're waiting anyways for his PR

src/client.rs
alx365: This code doesn't compile and i'm unsure on how to make it compile. mapping the option doesn't work because it uses data not owned by the current function.. i tried putting it in a cow (pun intended), but i can't do fn calls in match cases..

alx365:
> This definitely still needs some work. I think we should also generally wait until #45 is done, as that will affect this PR a lot.

changed all the things you asked for, now we just have to wait for druskus to finish

elkowar:
Now that #45  has been merged, you should be able to finish this @alx365

legendofmiracles:
I'll be rewriting this..

PR #73, opened 2020-12-13
Added support for custom config dir path

### LINKED ISSUE #67: [FEATURE] A -c flag to specify config location.
Just a -c flag to specify config location, e.g `eww open main_window -c ~/Documents/eww.xml` 

```
 `eww open main_window -c ~/Documents/eww.xml` 
```
# Additional context
just like polybar does

druskus20:
## Description

This PR implements support for a user specified config directory path. It aims to fix #67.

## Usage

Simply run the program with the arguments `-c [config_dir_path]` or `--config [config_dir_path]`

## Checklist

## TODO: 
- Update documentation
- I think the way server.rs handles configuration paths could be cleaned up a bit.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -54,14 +54,12 @@ fn main() {
             opts::Action::WithServer(action) => {
                 log::info!("Trying to find server process");
                 if let Ok(stream) = net::UnixStream::connect(&*IPC_SOCKET_PATH) {
-
                     client::forward_command_to_server(stream, action).context("Error while forwarding command to server")?;
                 } else if action.needs_server_running() {
                     println!("No eww server running");
-
                 } else {
                     log::info!("No server running, initializing server...");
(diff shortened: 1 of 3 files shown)


elkowar:
The code, as it currently is, wouldn't be able to properly deal with me _not_ specifying a config path, no? if no `-c` is specified, it should continue to default to `~/.config/eww/`

druskus20:
It should fallback to default which in this case would be CONFIG_DIR. Take a look at server.rs lines 24 to 30.
Here

alx365:
merge this before the edit subcommand (if you are still planning on merging it)

elkowar:
Still to do: not use empty string, but set a fallback value properly directly or at least use `Option<T>`
(first option would be to set the actual config path in the structopt macro, not sure if that's possible).
@druskus20 u gonna finish this? :P

PR #78, opened 2020-12-26
Add struts support and 'sticky' config option.

IanPoggers:
## Description

I added strut support which is configurable with the strut element in a window.

I changed the behavior of the 'focusable' attribute, which used to make the window a 'Popup' and set override redirect. The issue with both of these is they essentially tell the window manager to ignore the window and not manage it at all or check its properties. I think this is generally not desirable, and making the window a 'Dock' is sufficient. I've made 'popup' into its own option if people want the WM to ignore their widget. So setting `focusable="false"` should make it so you can't drag the window around, but the WM will still pay attention to properties like struts and stuff.

I also added a 'sticky' attribute to the window, which independently controls whether a  widget is visible on all workspaces. I have this set to true by default, but maybe this should be changed to better suit common use cases. Previously, if an eww window was not focusable, it would not be present on all workspaces, and now it would by default. This might mean some people need to change their configs.

## Usage

Simple example configuration:

```xml
<eww>
   <windows>
      <window name="main_window" focusable="false" sticky="true">
        <geometry anchor="top left" x="10px" y="10px" width="1900" height="30" />
        <struts top="50" top_end_x="1920" />
        <widget>
          <box>
            Cool eww bar
          </box>
        </widget>
      </window>
  </windows>
</eww>
```

The 'struts' element controls struts. By default everything is set to zero. Possible attributes are: left, right, top, bottom, left_start_y, left_end_y, right_start_y, right_end_y, top_start_x, top_end_x, bottom_start_x, bottom_end_x.

You should technically set the 'start' and 'end' of a strut, but I don't think most window managers properly support partial struts. At least awesome and openbox don't care from what I've noticed.

### Showcase

https://user-images.githubusercontent.com/68155222/103149510-05b15c80-471f-11eb-9acc-0482e25bf8c6.mp4

## Additional Notes

Struts won't compile on macos and the sticky feature doesn't seem to be working either.

Also struts behave kind of weird on some WM's, like on i3 the window resizes to fill the whole strut regardless of where I try to position it. But nothing I can really do about that.

## Checklist

- 

### DIFF
--- a/docs/content/main/configuration.md
+++ b/docs/content/main/configuration.md
@@ -216,6 +216,8 @@ The `<windows>` config should look something like this:
     </window>
 </windows>
 ```
+
+The window block contains multiple elements to configure the window.
 - `<geometry>` is used to specify the position and size of the window.
 - `<widget>` will contain the widget that is shown in the window.
 - `<struts>` Tells the window manager to reserve space at the edge of the screen (for a status bar or panel sort of widget).
--- a/docs/content/main/configuration.md
+++ b/docs/content/main/configuration.md
@@ -239,3 +239,6 @@ There are a couple things you can optionally configure on the window itself:
   This can be any number, representing the index of your monitor.
 - `sticky`: Controls whether the window is present on all workspaces.
   Possible values: `"true"`, `"false"`. Default: `"true"`
+- `popup`: Whether the window should be a popup. This basically makes the window manager ignore the window and not manage it.
+  Setting this to true will prevent struts from working, and it will be visible in all workspaces regardless of what 'sticky' is set to (because the WM is ignoring the window).
+  Possible values: `"true"`, `"false"`. Default: `"false"`
--- a/docs/content/main/configuration.md
+++ b/docs/content/main/configuration.md
@@ -225,8 +225,11 @@ The window block contains multiple elements to configure the window.
   The `left_start_y`, `left_end_y`, `right_start_y`, `right_end_y`, `top_start_x`, `top_end_x`, `bottom_start_x` and `bottom_end_x`
   attributes tell the WM the x or y limits of the strut. For example, `<struts top="30" top_start_x="0" top_end_x="960" />` would
   reserve a strut 30px down from the top edge of the screen, up to the 960th pixel from the left side of the screen.
-  Not all window managers support partial struts and may ignore the 'end' or 'start' attributes. All of these values default to zero if unset.
-  Also note that struts do not automatically position the window within the strut, it just tells the window manager you want to reserve space there.
+  Not all window managers support partial struts and may ignore the 'end' or 'start' attributes. All of these values default to zero if unset,
+  and it's not necessary to set all of them, only the ones that you need. You are also able to create multiple struts
+  if you'd like, like one at the top of the screen and another at the bottom. Also note that struts do not 
+  automatically position the window within the strut, it just tells the window manager you want to reserve space there.
+  [For more information check the X11 spec on window struts.](https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45760170756000)
 
 There are a couple things you can optionally configure on the window itself:
 - `stacking`: stacking describes if the window will be shown in the foreground (in front of other windows)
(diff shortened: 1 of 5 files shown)


docs/content/main/configuration.md
```diff
@@ -220,6 +220,13 @@ The `<windows>` config should look something like this:
 The window block contains multiple elements to configure the window.
 - `<geometry>` is used to specify the position and size of the window.
 - `<widget>` will contain the widget that is shown in the window.
+- `<struts>` Tells the window manager to reserve space at the edge of the screen (for a status bar or panel sort of widget).
+  The `left`, `right`, `top` and `bottom` attributes tell the WM how far to set struts away from that edge of the screen
+  The `left_start_y`, `left_end_y`, `right_start_y`, `right_end_y`, `top_start_x`, `top_end_x`, `bottom_start_x` and `bottom_end_x`
```
elkowar, line 225: It would be good to mention that it is not required to set all of these, and to explain how only setting for example top and bottom would behave, etc

src/app.rs
```diff
@@ -44,6 +50,53 @@ impl EwwWindow {
...
     }
+
+    #[cfg(not(target_os = "macos"))]
+    pub fn set_struts(&self) -> Result<()> {
+        let win: X11Window = self
+            .gtk_window
+            .get_window()
+            .expect("Could not get gdk window from gtk window")
+            .downcast()
+            .expect("Could not get X11 window from gdk window");
```
elkowar, line 61:
```suggestion
            .context("Could not get gdk window from gtk window")?
            .downcast()
            .context("Could not get X11 window from gdk window")?;
```

src/app.rs
```diff
@@ -44,6 +50,53 @@ impl EwwWindow {
...
+            def_struts.right_start_y,
+            def_struts.right_end_y,
+            def_struts.top_start_x,
+            def_struts.top_end_x,
+            def_struts.bottom_start_x,
+            def_struts.bottom_end_x,
+        ];
+
+        let (conn, _) = RustConnection::connect(None)?;
+        let get_atom = |name: &str| -> Result<_> { Ok(intern_atom(&conn, false, name.as_bytes())?.reply()?.atom) };
```
elkowar, line 81: This should just be a free, top-level function. Having it as a closure defined locally makes it seem more magic than it really is.

src/app.rs
```diff
@@ -44,6 +50,53 @@ impl EwwWindow {
     pub fn close(self) {
         self.gtk_window.close();
     }
+
+    #[cfg(not(target_os = "macos"))]
+    pub fn set_struts(&self) -> Result<()> {
```
elkowar, line 55: This should probably not be a method of `EwwWindow`, but instead be another free function that _takes_ `EwwWindow`, and is then used within `initialize_window`. Otherwise it would imply that it could make sense to call this multiple times, which really it doesn't - as setting struts is only ever part of the initial window initialization

docs/content/main/configuration.md
elkowar: also, a link to the X documentation describing how the _NET_WM_STRUT(_PARTIAL) works would probably be a good idea

elkowar (changes requested):
a couple small things, but BIG thanks already!

src/app.rs
```diff
@@ -372,3 +327,56 @@ fn get_monitor_geometry(n: i32) -> gdk::Rectangle {
...
+
+#[cfg(not(target_os = "macos"))]
+fn set_struts(eww_window: &EwwWindow) -> Result<()> {
+    let x11win: X11Window = eww_window
+        .gtk_window
+        .get_window()
+        .context("Could not get GDK window from GTK window")?
+        .downcast()
+        .ok()
+        .context("Could not get X11 widnow from GDK window")?;
```
elkowar, line 339: there is typo in `widnow`

elkowar (changes requested):
Looks great, except for typo :D

elkowar:
Closing this, as it's been implemented in https://github.com/elkowar/eww/commit/59906306772dd1cc0875bccf6c033e13624b4d85 in the gtk4 branch.

Thanks a _lot_ for the work you've done here, I was able to copy over most of the knowledge and a bit of the code you came up with here. Thanks!

PR #82, opened 2021-01-01
Fix #81

### LINKED ISSUE #81: [BUG] SCSS variables get replaced with environment variables
## Describe the bug
It's impossible to use [SCSS variables](https://sass-lang.com/documentation/variables) in `eww.scss`. The way that `eww` parses the SCSS files means that any tokens that look like `$myvariable` get swapped with an environment variable. If the environment variable does not exist, the token gets swapped with a null value, producing invalid SCSS and a null stylesheet.

## Reproducing the issue
Specify a SCSS variable in `eww.scss`. Using a variable name that isn't specified in the environment generates an invalid `scss` file and causes the widget to be unstyled. 

```scss
// eww.scss
window {
  background: #121212;
}

$red: #AE2955;
.red {
  color: $red;
}

scale trough {
  all: unset;
  min-height: 5px;
  border-radius: 50px;
  background-color: darken($red, 20%);
  min-width: 150px;
}
highlight {
  all: unset;
  background-color: red;
  background-color: #ae2955;
  color: rgba(0,0,0,0);
}
slider {
  all: unset;
}
```

```html
<eww>
  <definitions>
    <def name="controls">
      <box class="sidebarContainer controls" orientation="v">
        <box class="red" space-evenly="false" halign="center">
          <box class="icon">
            ☼
          </box>
          <scale min="0" max="100" value="10"></scale>
        </box>
      </box>
    </def>
  </definitions>


  <windows>
    <window name="controls_side">
      <geometry x="50px" y="210px" height="100px" width="250px"/>
      <widget>
        <controls />
      </widget>
    </window>
  </windows>
</eww>

```

## Expected behaviour
Display proper styles while using the specified SCSS variable.

#### Current behavior
image
#### Expected behavior
image

### Possible solutions

1. Swapping environment variables after the SCSS->CSS conversion
   - Unfortunately this would not work because any undeclared environment variables would throw an error in the SCSS->CSS conversion step
2. Change the environment variable syntax as to not conflict with SCSS (e.g. `${MY_ENV_VAR}`)
   - Would break existing configs that use environment variables, but would prevent a collision with existing SCSS syntax.
3. Leave `$myvar` strings alone when there is not an environment variable to swap it with
   - The current behavior swaps undefined `$variables` with an empty string. It feels like we could happily leave these variables alone and win SCSS variables, but we could run into a situation where there's a name collision with a SCSS variable and we're back at square one; fails to compile when `$CWD: #000` gets turned into `/home/user/: #000`... contrived, but possible. 
4. Intelligently create a list of SCSS variables and don't swap these when we're looking for environment variables
  - [grass's](https://github.com/connorskees/grass) API is not that open, I don't think this is reasonable to do. 

I'm most okay with option 3 if there's a more verbose error message that tells me why my SCSS failed to compile. That allows the user to resolve the issue with their SCSS file.

Any other solutions?

jack-michaud:
Fix for #81 

## Description

Instead of replacing all SCSS variables that do not have a non-null environment variable with an empty string, leave the SCSS variable alone. This is option 3 described in #81. 

Adds error message when trying to load invalid `scss` file

e.g. putting in `$PWD: #000;` in the place of `$red` variable declaration from example in #81, 
```
jack@CASTOR ~> eww open controls_side
error parsing scss file: encountered SCSS parsing error: SassError { kind: ParseError { message: "Undefined variable.", loc: SpanLoc { file: File("stdin"), begin: LineCol { line: 6, column: 9 }, end: LineCol { line: 6, column: 14 } }, unicode: true } }
```
which should allow the user to diagnose the issue (though if a less cryptic message is desired, we can try to elucidate if possible).

## Usage

See example from #81. Intended behavior should be output.

## Additional Notes

This is WIP because this is one of the proposed solutions, but it is the one that made the most sense to me and I implemented it for my own setup. This could be a good starting point for the real solution.

## Checklist

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -49,8 +49,13 @@ pub fn initialize_server(should_detach: bool, action: opts::ActionWithServer) ->
         gtk::StyleContext::add_provider_for_screen(&screen, &app.css_provider, gtk::STYLE_PROVIDER_PRIORITY_APPLICATION);
     }
 
-    if let Ok(eww_css) = util::parse_scss_from_file(&scss_file_path) {
-        app.load_css(&eww_css)?;
+    match util::parse_scss_from_file(&scss_file_path) {
+        Err(err) => {
+            eprintln!("error parsing scss file: {}", err);
+        },
+        Ok(eww_css) => {
+            app.load_css(&eww_css)?;
+        }
     }
 
     // run the command that eww was started with
--- a/src/server.rs
+++ b/src/server.rs
@@ -52,7 +52,7 @@ pub fn initialize_server(should_detach: bool, action: opts::ActionWithServer) ->
     match util::parse_scss_from_file(&scss_file_path) {
         Err(err) => {
             eprintln!("error parsing scss file: {}", err);
-        },
+        }
         Ok(eww_css) => {
             app.load_css(&eww_css)?;
         }
--- a/src/util.rs
+++ b/src/util.rs
@@ -77,15 +77,16 @@ pub fn parse_duration(s: &str) -> Result<std::time::Duration> {
 }
 
 /// Replace all env-var references of the format `"something $foo"` in a string
-/// by the actual env-variables. If the env-var isn't found, will replace the
-/// reference with an empty string.
+/// by the actual env-variables. If the env-var isn't found, the variable
+/// will be left alone.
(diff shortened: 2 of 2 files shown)


elkowar:
See my comment on #81

Issue #56, opened 2020-11-06
Provide example configurations


elkowar:
now that documentation is mostly taken care of, the main thing missing to make eww approachable is good examples. These examples should include both the css and xml and explain how and why things are used the way they are. It should also be shown how a script can be used to change the eww-state from outside. The use of the literal tag as well as script vars (both interval and tail) should be included. 

These examples may in the future be used in the CI-pipeline as a means of veryifying that changes don't brake existing functionality.

## acceptance criteria
- example widget config fulfilling the criteria stated above is implemented.
- screenshots showing the widgets are provided.
- examples are mentioned in the appropriate places in the documentation.

elkowar:
@Swexti would you be willing to look into this? just writing a couple minimal, functional examples that showcase the functionality and configuration of eww

Swexti:
uh, idk, how in-depth do i have to go with "explain how and why things are used the way they are"? Like a tutorial? Or just a short description of what it does?
so e.g:

```
If we take a look at this script we can see that once it gets the weather information (shown by the `weather=$("curl wttr.in")`), it tries to find what `if` tag equals the weather. So if the weather is Rain and we have two if fuctions that look like this:
`
if [ $weather = 'Sun']; then;
   eww update weather Sunny
elif [ $weather = 'Rain';] then;
   eww update weather Rain
`
It matches the secon if tag (the elif) with 'Rain' since the $weather = Rain. And since the $weather isn't Sunny, it doesn't match with `$weather = 'Sun'

After that it updates the `<script-var>` to match what weather information it gave with `eww update weather <weather>`
```
or something like
```
This script updates the `<script-var>` once the weather information matches the `if` tag.
```

elkowar:
just some actual, full configurations with a couple short comments that explain the non-obvious stuff. so more like version 2 :D

Swexti:
> non-obvious stuff

ah okay i thought i had to explain everything. But i don't have to do like `This is the <image> tag, it displays an image with the specified width and height`? Since the widget documentation takes care of that?

elkowar:
yea, exactly. no need to explain everything, it's mostly just about having examples. just write some stuff and i'll tell u where some comments may be good

Swexti:
ah okay

elkowar:
perfect, I'll assign you!

6gk:
bruhhh, i know this is an example, but still

> ```
> if [ $weather = 'Sun']; then;
>    eww update weather Sunny
[...]

PR #99, opened 2021-01-20
Provide an example configuration (with space for more!)

Axarva:
## Description
Provides example configurations (#56), also fixes the infamous `informatiion` typo.

## Usage
Put files from the `examples/` directory **recursively** to your `~/.config/eww/` directory and launch the `bar` window.

### Showcase
I know, I know, showed this already but yeah

image

## Additional Notes
Added a line in the docs as well.

## Checklist


elkowar:
I think it would be nice if the different modules in the bar where sourced out into separate widget definitions, that are then all called from the main bar widget. 
As in, specify a `<def name="workspaces>` and include that in the `bar` definition, and so on - to show off how widgets can be nested and reused.

examples/scripts/getram
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
elkowar, line 1:
```suggestion
#!/usr/bin/env bash
```

examples/scripts/getvol
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
elkowar, line 1:
```suggestion
#!/bin/sh
```

elkowar:
also, it would be cool if you could add the screenshot to the Readme and the documentation too, maybe.

Axarva:
> I think it would be nice if the different modules in the bar were sourced out into separate widget definitions, that are then all called from the main bar widget.
> As in, specify a `<def name="workspaces">` and include that in the `bar` definition, and so on - to show off how widgets can be nested and reused.

Did you mean to make multiple definitions for widgets and then include them all in the same `window`? 
As in different defs for everything and then include them all in the same window
I agree that would modularize stuff a lot but I think I tried that and it didn't work out due to widget having more than one child.
image

Axarva:
> also, it would be cool if you could add the screenshot to the Readme and the documentation too, maybe.

Did it all, I wasn't somehow notified yesterday, so it took time, but yes, everything is done now lol

examples/scripts/getram
Axarva: Just a habit lol

elkowar:
> Did you mean to make multiple definitions for widgets and then include them all in the same window?

you can use user-defined widgets in other user-defined widgets. So you can create a `<time>` widget you can then use in a bar widget, which is then used in the bar window.

Axarva:
Will try that and see how it works out :)

Axarva:
@elkowar Does it now match what you had in mind?

elkowar:
Yea, this is clean now! Thanks!

PR #105, opened 2021-01-30
Shows warnings when defining variables that are already defined, and other general improvements

legendofmiracles:
## Description

This PR adds support for warnings when the same variables/windows/widgets and initial script variables in different config files.  
And other general improvements.

## Usage

Include another config, define something, which is defined in the main config.

The warnings will then be shown in `eww logs`

### Showcase

```
The normal config, and the included config: /home/legend/.config/eww/test.xml, both define a widget, with the same name: 'combo-box'
```

## Checklist

- [x] I used `cargo fmt` to automatically format all code before

### DIFF
--- a/docs/content/main/configuration.md
+++ b/docs/content/main/configuration.md
@@ -118,10 +118,12 @@ Here you can include other config files so that they are merged together at star
 
 ```xml
 <includes>
-  <file path="./something.xml"/>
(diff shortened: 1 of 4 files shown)


legendofmiracles (commented):
Looks good 👍

PR #117, opened 2021-02-21
Predefined variables/Magic vars

legendofmiracles:
## Description
Adds always there variables, that can be used without any extra work. 
Also reworks the gen docs script a bit. 
And eww-state only displays used variables, to stop polluting the output with all the unused magic vars.
## Usage
An example of RAM usage would be:
`{{EWW_RAM}}`, which can be used anywhere, normal vars would work.

### Showcase
image
network usage while downloading a game: 
image

## Additional Notes

The reason why there are several files, for stuff like network up and down, is because there would be a data race, of both updating, thus giving false values. And to make a system, that updates once, and then gives the value you want, would be very complex. 

## Checklist

- [x] All things I've added are correctly documented.


Axarva:
So... these variables are predefined within eww so we don’t need to say, write scripts such as “cat /sys/power/...” to get the battery level or something? Would these work on MacOS as well?

legendofmiracles:
Most of them should work on macos, haven't tested it tho.

elkowar:
I think it'd be nicer to have the separate data sources share a single SYSTEM thingy that has all of the features, rather than each having their own (the builder API should allow for adding all of the `with_something` thingys, no?)
Then you can also have them all in one module (-> all in one file), which i'd prefer - this is _a lot_ of files

src/config/system_stats/battery.rs
```diff
@@ -0,0 +1,41 @@
...
+    Ok(number.parse().context("Couldn't make a number from the parsed text")?)
+}
+
+fn linux() -> Result<u8> {
+    Ok(std::fs::read_to_string("/sys/class/power_supply/BAT0/capacity")
+        .context("Couldn't get battery info from /sys/class/power_supply/BAT0/capacity")?
+        .trim()
+        .parse()
+        .context("Couldn't parse the number in /sys/class/power_supply/BAT0/capacity")?)
+}
```
elkowar, line 41: just have a function per os, like
```rs

#[cfg(target_os = "linux")]
fn get_battery_capacity() -> Result<u8> {
   Ok(std::fs::read_to_string("/sys/class/power_supply/BAT0/capacity")
        .context("Couldn't get battery info from /sys/class/power_supply/BAT0/capacity")?
        .trim()
        .parse()
        .context("Couldn't parse the number in /sys/class/power_supply/BAT0/capacity")?)
}

#[cfg(target_os = "macos")]
pub fn get_battery_capacity() -> Result<u8> {
    use regex::Regex;
    let capacity = String::from_utf8(
        std::process::Command::new("pmset")
            .args(&["-g", "batt"])
            .output()
            .context("\nError while getting the battery value on macos, with `pmset`: ")?
            .stdout,
    )?;
    let regex = Regex::new(r"[0-9]*%")?;
    let mut number = regex.captures(&capacity).unwrap().get(0).unwrap().as_str().to_string();
    number.pop();
    Ok(number.parse().context("Couldn't make a number from the parsed text")?)
  }
```

src/config/system_stats/battery.rs
```diff
@@ -0,0 +1,41 @@
...
+    let capacity = String::from_utf8(
+        std::process::Command::new("pmset")
+            .args(&["-g", "batt"])
+            .output()
+            .context("\nError while getting the battery value on macos, with `pmset`: ")?
+            .stdout,
+    )?;
+    let regex = Regex::new(r"[0-9]*%")?;
+    let mut number = regex.captures(&capacity).unwrap().get(0).unwrap().as_str().to_string();
+    number.pop();
```
elkowar, line 31: this is,... really confusing.
what does number.pop even do here, remove a character? why? what is that character?
In this case, please explain stuff more/make code more obvious (if it's stripping whitespace, use strip/trim or whatever exists instead), and provide a comment that showcases an example of the parsed data, so it's clear what/how data looks.

src/config/system_stats/ram.rs
```diff
@@ -0,0 +1,14 @@
...
+
+lazy_static! {
+    static ref SYSTEM: Arc<Mutex<System>> = Arc::new(Mutex::new(System::new_with_specifics(RefreshKind::new().with_memory())));
+}
+
+pub fn ram() -> f32 {
+    let sys = SYSTEM.clone();
+    let mut c = sys.lock().unwrap();
+    c.refresh_memory();
+    (c.get_used_memory() as f32 + c.get_used_swap() as f32) / 1000000 as f32
```
elkowar, line 13:
```suggestion
    (c.get_used_memory() as f32 + c.get_used_swap() as f32) / 1_000_000f32
```
should work

src/config/system_stats/network_down.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+pub fn get_down() -> f32 {
+    let sys = SYSTEM.clone();
+    let mut c = sys.lock().unwrap();
+    c.refresh_networks();
+    let interfaces: u64 = c
+        .get_networks()
+        .iter()
+        .filter(|a| !a.0.starts_with("docker") || !a.0.starts_with("lo"))
```
elkowar, line 18: bruh please tell me there is a way to do this more cleanly,... like some way to detect internal (logical?) interfaces generically,... this is really ewwwwwwwww

src/config/system_stats/disk.rs
```diff
@@ -0,0 +1,22 @@
...
+    let sys = SYSTEM.clone();
+    let mut c = sys.lock().unwrap();
+    c.refresh_disks_list();
+
+    let root = c
+        .get_disks()
+        .iter()
+        .find(|&x| x.get_mount_point() == std::path::Path::new("/"))
+        .ok_or_else(|| anyhow!("Couldn't find a drive mounted at /"))?;
+    Ok((root.get_total_space() as f32 - root.get_available_space() as f32) / 1000000000 as f32)
```
elkowar, line 21:
```suggestion
    Ok((root.get_total_space() as f32 - root.get_available_space() as f32) / 1_000_000_000f32)
```

gen-docs.js
```diff
@@ -1,12 +1,28 @@
...
+    for (let line of data.split("\n")) {
+        let match = line.match(pattern);
+        if (match) {
+            let split = match[1].split("-");
+            let name = split[0].trim();
+            let desc = split[1].trim();
+            stream.write(`### \`${name}\`\n${desc}\n`);
+        }
+    }
+}
```
elkowar, line 25: Not really a fan of the stream stuff here, as it seems,.... definitely unnecessarily complex for no real benefit.

Also, `printDocs` should be renamed to something more fitting if it now writes to the widgets file directly (which i'm also not really a fan of, as that means that the script is a lot more anoying to test now .-.)

elkowar: I'd argue in favour of just making a generateVarDoc function that does all of the variable documentation generation, and a generateWidgetDoc function (doing the same for the widget stuff), and then maybe just writing those to the respective files directly. Then i'd also be fine with writing directly to the files i guess

elkowar (changes requested):
Some general things, but all in all, _big_ thanks_ already!

Specifically, i think that the modules for data sourcing don't really need to be as split-up into different files as they are, but instead should just share a single SYSTEM global and all be in a single file. Also, i don't think you need to `.clone` the system all the time, a `*SYSTEM.lock()` or sth like that  _should_ work (?).

Also, the formatting in some places is really weird, please make sure to run everything through the autoformatter

src/config/inbuilt.rs
```diff
@@ -0,0 +1,80 @@
+// Network and CPU load have to be done with the *async magic*
+use crate::{
+    config::{system_stats::*, PollScriptVar, ScriptVar, VarSource::Function},
+    value::{PrimitiveValue, VarName},
+};
+use anyhow::*;
+use std::{collections::HashMap, time::Duration};
+
+pub fn get_inbuilt_vars() -> HashMap<VarName, ScriptVar> {
```
elkowar, line 9: This function has some formatting issues, but even if they are fixed it's only semiiii readable.
- i don't think you need to provide the return type in the lambda explicitly, but if you do, then you at least don't need to specify `Error` (i think), as it should know that it's talking about `anyhow::Result` which only needs one value.
- i'm not a fan of importing `VarSource::Function` globally. Either don't at all, and use `VarSource::Function` explicitly (which i'd likely be in favour of, as long as it doesn't add additional line wrapping), or use a local use
(you can use the `use` statement inside of `get_inbuilt_vars`)

src/config/system_stats/battery.rs
legendofmiracles: This pops of the % of the output from the command.

src/config/system_stats/network_down.rs
legendofmiracles: well that's the entire point, is it not? 
For real tho, `route` can display which interface is being used to communicate with the internet. I can look into this. No promises tho

src/config/inbuilt.rs
elkowar: or in fact, not sure if this is a great idea, but what do you think about using this:
```rs
macro_rules! builtin_vars {
    ($interval:expr, $($name:literal => $fun:expr),*$(,)?) => {{
        hashmap! {
            $(
            VarName::from($name) => ScriptVar::Poll(PollScriptVar {
                name: VarName::from($name),
                command: VarSource::Function($fun),
                interval: $interval,
            })
            ),*
        }
    }}
}
```
which could result in sth like
```
let data = builtin_vars! { std::time::Duration::from_seconds(1000),
  "EWW_CORES" => || Ok(PrimitiveValue::from(format("{:.1}", cores::cores()))),
  "EWW_BATTERY" => || Ok(PrimitiveValue::from(
    match battery::get_battery_capacity() {
        Err(e) => {log::error!("Couldn't get the battery capacity: {:?}", e); f32::NAN }
        Ok(o) => o as f32,
    }
  ))
};
 ```
 not 100% sure if i like this, but it could definitely clean up the code a lot,.... what do you think?

Comment (elkowar) [reply src/config/inbuilt.rs:9]: image
expands very sanely, too 👍🏻 

(jk, but this is not an issue, it's just weirdly formatted)

File: src/config/system_stats/battery.rs (hunk 2 above)
Comment (elkowar) [reply src/config/system_stats/battery.rs]: yea, please do add a comment explaining that (optimally show an example of the raw data that is being parsed)

File: gen-docs.js (hunk 6 above)
Comment (legendofmiracles) [reply gen-docs.js:25]: Was thinking of doing that, but i didn't want to spend any more time in js, than necessary ;-;
My way of debugging has been generating the docs and then just `git restore docs/...` to the original state, without generated content. 
And i didn't really want to touch your js code, as i didn't really understand it or had the desire too -.-

File: src/config/inbuilt.rs (hunk 7 above)
Comment (legendofmiracles) [reply src/config/inbuilt.rs:9]: Yeah this looks great. 
I'll use your macro, thanks!

File: src/config/system_stats/network_down.rs (hunk 4 above)
Comment (legendofmiracles) [reply src/config/system_stats/network_down.rs:18]: It seems like /proc/self/net/route has all of the routes, and by that extension also only all interfaces that are connected to the internet, or a network large enough, where this would be necessary.

So i could parse this and then only use the interfaces found in that file.

**Comment by legendofmiracles** [comment]
The issue with one system thing would be, that no new vars could be added at runtime, currently, because of the bug that every magic var is executed once, that'd be no problem, but later one it should be hot reload-able
and certain things like net up and down need different ones, or there'd have to be a system in place to refresh once, and only get the one you want, while also supporting to get both.

File: gen-docs.js (hunk 6 above)
Comment (elkowar) [reply gen-docs.js:25]: to be honest, just get rid of the streaming stuff (really not necessary) and maybe make the naming a bit more consistent - I've been planning to rework the doc script completely anyways (I want to use it to generate docs for everything, adding some more flexibility by allowing for a more complex widget structure - I really dislike having any specific documentation for xml-tags/properties not be synced with the code, so I'm going to try to have nearly _all_ of the docs be generated from comments in the code)

File: src/config/system_stats/network_down.rs (hunk 4 above)
Comment (elkowar) [reply src/config/system_stats/network_down.rs]: if that's the best way, then sure. I'd have hoped that the library provides some direct functionality for that, but if not then so be it, parsing that file sounds like a good alternative. Make sure that you keep macos compat in mind though (i'd be _fine_ with having _less_ functionality on macos - i.e. by just not checking for only active network interfaces there or whatever (although doing the same checks for macos would be optimal, of course), but it's important that it at least doesn't crash :D)

**Comment by elkowar** [comment]
> The issue with one system thing would be, that no new vars could be added at runtime, currently, because of the bug that every magic var is executed once, that'd be no problem, but later one it should be hot reload-able
> and certain things like net up and down need different ones, or there'd have to be a system in place to refresh once, and only get the one you want, while also supporting to get both.

So from how I understand the docs of sysinfo, that shouldn't be an issue:
`System::new()` gives you a `System` struct with no data loaded. The system-struct represents the current state of the System - it contains a field for each of the stats (i.e. cpu load,  boot time, kernel version, uptime, and so on).

the `System::refresh_SOMETHING(&mut self)` methods then load the desired data and store it in the struct. Using `System::new_with_specifics` just means that it will load the respective bits of data once, initially. Thus, creating a global `System` with `System::new()` and then, within each of the data-fetching thingys, just `refresh_mem-or-whatever`-ing the respective bits of data should work without issues, and thus remove the need for unnecessary memory usage and unnecessary amounts of global variables.

Alternatively it would be possible to just create a new `System` with `System::new_with_specifics` each time, which would probably be more than fine in practice, but feels worse (as it allocates a lot of unused data in that case, similarly to how it does now), so I'd do the global `System`.

**Comment by legendofmiracles** [comment]
Ok i can do that, but an issue would be, that net up and down are reading from the same source, while refreshing at the same time, thus making all numbers it returns useless. So i'd also need some `Mutex<bool>`, if another one is running or not, to then not refresh and let the other one do that for you.

**Comment by undefinedDarkness** [comment]
:confused:
wait this isnt just using system commands like a fetch script?

**Comment by legendofmiracles** [comment]
No, it's mostly using libraries. As that's always much nicer than depending on binaries being present, and parsing output from programs that, who's output can always change.

**Comment by undefinedDarkness** [comment]
Stupid question but would it be possible to have this under a feature flag?
That might not be necessary tho :thinking:

**Comment by elkowar** [comment]
> Stupid question but would it be possible to have this under a feature flag?
> That might not be necessary tho

could be, but i don't really see the point. this is comparatively _very_ simple functionality with very little overhead (both runtime and compile-time)

**Comment by legendofmiracles** [comment]
If you don't use these variables, they don't have any impact at all, only reading a bit of data that's baked into the binary, on daemon startup.

**Comment by undefinedDarkness** [comment]
On in that case, nvm then

**Comment by legendofmiracles** [comment]
Sorry for so many commits ;-;
I need to keep resolving the merge conflicts.

**Comment by elkowar** [comment]
(quick trick for the rebase, as i renamed `AttrValue` to `AttrVal` and that'd otherwise be _super_ painful:
use your editor to rename the `AttrVal` back to `AttrValue`, then rename back again. that way it'll pick up your changes)

**Comment by elkowar** [comment]
plox resolve merge conflicts uwu

**Comment by legendofmiracles** [comment]
@elkowar done - should i squash some of the commits?

**Comment by elkowar** [comment]
Why is that? Lmao

**Comment by legendofmiracles** [comment]
Instead of pulling your changes i merged onto master, so now rebasing back 28 commits (the amount of commits that github claims) I also see all other commits that happened in that time, not on my branch.

**Comment by elkowar** [comment]
Oh god
_git_ good my friend, **git** good.

Its fine tho, i'll just squash everything when i merge ig. I'll do some final testing today, if everything goes well i'll merge

File: docs/content/main/magic-vars.md [hunk 8]
```diff
@@ -0,0 +1,14 @@
...
++++
+
+# Magic Variables
+These are variables that are always there, without you having to import them.
+
+The delay between the updating variables is always 2s.
+
+
+## Updating magic variables
+
```
elkowar, line 14: This file is,... missing stuff 👀 specifically, missing the actual variables - and idk what `updating magic variables` was even supposed to be

elkowar: Ahhhhhh that's getting autogenerated, nevermind 👀

elkowar:
`EWW_CORES` is `NaN` for me 👀 that shouldn't be the case, no? 
(note: AMD Ryzen 3700X, if that matters)
Also, i'd argue the name is kinda confusing, sth like `EWW_CPU_TEMP` would probably be more fitting

src/config/inbuilt.rs
```diff
@@ -0,0 +1,53 @@
...
+            }
+        ))),
+
+        // @desc EWW_BATTERY - Battery capacity in procent of the main battery
+        "EWW_BATTERY" => || Ok(PrimitiveValue::from(
+            match get_battery_capacity() {
+                Err(e) => {log::error!("Couldn't get the battery capacity: {:?}", e); f32::NAN }
+                Ok(o) => o as f32,
+            }
+        )),
```
elkowar, line 42: Formatting 🙏🏻 
```suggestion
        // @desc EWW_DISK - Used space on / in GB (Might report inaccurately on some filesystems, like btrfs)
        "EWW_DISK" => || Ok(PrimitiveValue::from(format!("{:.1}",
            match disk() {
                Err(e) => {
                    log::error!("Couldn't get disk usage on `/`: {:?}", e);
                    f32::NAN
                }
                Ok(o) => o
            }
        ))),

        // @desc EWW_BATTERY - Battery capacity in procent of the main battery
        "EWW_BATTERY" => || Ok(PrimitiveValue::from(
            match get_battery_capacity() {
                Err(e) => {
                    log::error!("Couldn't get the battery capacity: {:?}", e);
                    f32::NAN
                }
                Ok(o) => o as f32,
            }
        )),

```

elkowar (changes requested):
Some more smol style things i just noticed... Sorry lol

elkowar:
> `EWW_CORES` is `NaN` for me  that shouldn't be the case, no?
> (note: AMD Ryzen 3700X, if that matters)
> Also, i'd argue the name is kinda confusing, sth like `EWW_CPU_TEMP` would probably be more fitting

@legendofmiracles anything for this? That's the last thing missing currently 👀

legendofmiracles:
I'll test it when I got the time to

legendofmiracles:
I'll try to make a minimal reproducible example, bla bla.
But currently i don't have the time to do that, maybe during the next week.

src/config/system_stats.rs
```diff
@@ -0,0 +1,123 @@
+use lazy_static::lazy_static;
+#[cfg(target_os = "macos")]
+use regex::Regex;
```
elkowar, line 3: Can we move this conditional `use`into the function that uses it?(just `use` in get_battery_capacity, rather than here)

elkowar:
ahhh yes, right,...
Have fun with gen-docs, lol. Has been rewritten now. I'm sorry ^^'

undefinedDarkness:
Most of its still there, I just slightly modified stuff I wasnt happy with

legendofmiracles:
I fixed it, but just straight copied and pasted what I programmed earlier, I really don't know TS, so someone should probably rewrite it in the ts way:tm:

legendofmiracles:
Also, network is a json value now

legendofmiracles:
Love the added/removed lines =D

elkowar:
image
This doesn't yet look all that good i'm afraid, lol. The doc-generation is definitely not done yet.

elkowar:
EWW_NET currently gives me 
```
EWW_NET: { "NET_DOWN": {}, "NET_UP": {}}
```
so something is definitely broken there

elkowar:
maybe you can make your fix be +420?

gen-docs.ts
```diff
@@ -12,6 +12,21 @@ interface Widget {
...
+    for (let line of data.split("\n")) {
+        let match = line.match(pattern);
+        if (match) {
+            let split = match[1].split("-");
+            let name = split[0].trim();
+            let desc = split[1].trim().replaceAll("\\n", "\n\n");
+            output = output + `### \`${name}\`\n${desc}\n`;
+        }
+    }
+    return output;
```
mlvzk, line 27:
```suggestion
    const pattern = /^.*\/\/\s*@desc\s*(\w+)\s*-\s*(.*)$/gm;
    let output = [];
    for (const [_, name, desc] of data.matchAll(pattern)) {
        output.push(
`### \`${name}\`
${desc.replaceAll("\\n", "\n\n")}
`);
    }
    return output.join("\n");
```

legendofmiracles:
> This doesn't yet look all that good i'm afraid, lol. The doc-generation is definitely not done yet.

@elkowar 
Is this better? 
image

elkowar:
> Used space on / in GB (Might report inaccurately on some filesystems, like btrfs)
Description isn't really accurate, as it contains more than that - please update to mention what it actually is

elkowar:
also still, rename to `EWW_CPU_TEMPS` plox, and then `EWW_CPU_USAGE` or sth
also, can we do per-core cpu usage on the CPU too? and then maybe add an additional `average` field?

like that `EWW_CPU_USAGE` is `{ "core1": 20, "core2": 30, total: "25" }`?

legendofmiracles:
I'm unsure how to handle the battery stuff since every computer seems to have a different directory structure, or maybe there are only two and we can check which one it is.. It's not really documented

Wouldn't it be sufficient, for now, to just give the user the `state` (charging, full, discharging) and the capacity (% full)?

elkowar:
Maybe look into how other projects are doing it? Not really a fan of providing that functionality but then limiting it to be only for battery  0 (if that's what you're implying)

legendofmiracles:
That's not what i meant, i meant only having the values of capacity and status for each battery

elkowar:
Ehhh that's better, but i'd still like to provide a total value too,.... again, have you checked how other projects do it?

legendofmiracles:
@elkowar i think this should work, couldn't really test this on systems with `energy_*`, so if someone who has said files in `/sys/class/power_supply/BAT<N>` that would be pretty pog
But in my manual tests by faking a battery it worked

(also should i remove one line somewhere?)

elkowar:
Given that it seems to work for everyone that tries it - would you say it's ready to be merged? And sure, if you wanna get that 420, do remove one line somewhere, lol

legendofmiracles:
I think it is, I'm also quite happy with the code if I may say so myself, it went a *long* way. Only thing i'm not happy with is the git workflow that i had throughout the time, but it's too late to change that now :/

Because of Cargo.lock merge conflicts it's a few too many lines longer than 420, not worth the effort.

elkowar:
Genuinely a really big feature, and a large improvement to eww. thanks a ton!

undefinedDarkness:
🥳

PR #124, opened 2021-03-01
Expression language

elkowar:
Adds the Eww expression language.
This means that now, within variable interpolations (`{{foo}}`) you can embed some expressions, i.e.
```
{{ if 12 > 15 + 3 then "hi" else some_var }}
```


undefinedDarkness:
I need this in my life

Issue #102, opened 2021-01-23
[BUG] [macOS] eww doesn't compile on macOS because inotify isn't a thing on mac


Rougemoot:
Pretty much just what it says in the title.

Maybe `fswatch` is the way to go to replace inotify, but I don't know what I'm talking about, really. I just wanted to write it down here after our conversation on discord.

legendofmiracles:
This should be fixed as inotify is now a linux only dependency.

Rougemoot:
Meaning no hot-reload on macOS?

mac-codes9:
Did anyone ever get it compiling on macOS? I keep running into errors and there isn't much documentation

PR #141, opened 2021-03-26
Hot reload

undefinedDarkness:
## Description
Uses the [notify](https://github.com/notify-rs/notify) crate to implement hot reloading & now eww does the equivalent of `eww reload` instead of what it used to do previously

## Usage
Open a text editor in eww's config dir & start editing stuff, idk

### Showcase
[Here](https://cdn.discordapp.com/attachments/635625917623828520/824167882098212894/Screencast_2021-03-24_115353.mp4)

## Additional Notes
Could probably be done in a better way hehe
But this works, ill change it if there are any problems

This also should help #102 since notify is a wrapper around winapi, fsevents & inotify for windows, macos & linux respectively
Eww on windows when??
Elkowar make eww ipc cross platform pls 🥺

## Checklist

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1,8 +1,11 @@
 use crate::{app, config, eww_state::*, ipc_server, script_var_handler, try_logging_errors, util, EwwPaths};
 use anyhow::*;
+use app::DaemonResponse;
 use futures_util::StreamExt;
 use std::{collections::HashMap, os::unix::io::AsRawFd, path::Path};
 use tokio::sync::mpsc::*;
+
+use notify::Watcher;
 
 pub fn initialize_server(paths: EwwPaths) -> Result<()> {
     do_detach(&paths.get_log_file())?;
@@ -70,12 +73,15 @@ pub fn initialize_server(paths: EwwPaths) -> Result<()> {
 
 fn init_async_part(paths: EwwPaths, ui_send: UnboundedSender<app::DaemonCommand>) {
     std::thread::spawn(move || {
-        let rt = tokio::runtime::Builder::new_multi_thread().enable_all().build().expect("Failed to initialize tokio runtime");
+        let rt = tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .build()
+            .expect("Failed to initialize tokio runtime");
         rt.block_on(async {
             let filewatch_join_handle = {
                 let ui_send = ui_send.clone();
                 let paths = paths.clone();
-                tokio::spawn(async move { run_filewatch(paths.get_eww_xml_path(), paths.get_eww_scss_path(), ui_send).await })
+                tokio::spawn(async move { run_filewatch(paths.config_dir, ui_send).await })
             };
 
             let ipc_server_join_handle = {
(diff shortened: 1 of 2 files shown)


elkowar:
Yea, currently, this just blocks one of the threads in the tokio threadpool - of which there may only be one, theoretically. Thus, this could very easily hang up the whole application, which is not really an option.

If we _really_ need to use actual threads here and there _reeeallllyyy_ isn't a solution that actually uses async properly, you should use `spawn_blocking` to spawn the run_filewatch task, such that that happens on a separate thread. Still ugly, but at least not a critical blocking issue

src/server.rs
```diff
@@ -103,52 +109,35 @@ fn init_async_part(paths: EwwPaths, ui_send: UnboundedSender<app::DaemonCommand>
...
-                        log::info!("Reloading eww configuration");
-                        let new_eww_config = config::EwwConfig::read_from_file(config_file_path.as_ref())?;
-                        evt_send.send(app::DaemonCommand::UpdateConfig(new_eww_config))?;
-                } else if event.wd == scss_file_descriptor {
-                        log::info!("reloading eww css file");
-                        let eww_css = crate::util::parse_scss_from_file(scss_file_path.as_ref())?;
-                        evt_send.send(app::DaemonCommand::UpdateCss(eww_css))?;
-                } else {
-                    eprintln!("Got inotify event for unknown thing: {:?}", event);
+async fn run_filewatch<P: AsRef<Path>>(config_dir: P, evt_send: UnboundedSender<app::DaemonCommand>) -> Result<()> {
```
elkowar, line 113: Slight cleanup for this function:
```rs
async fn run_filewatch<P: AsRef<Path>>(config_dir: P, evt_send: UnboundedSender<app::DaemonCommand>) -> Result<()> {
    use notify::*;

    let (watch_send, watch_recv) = std::sync::mpsc::channel();
    let mut watcher = watcher(watch_send, std::time::Duration::from_secs(0))?;
    watcher.watch(&config_dir, notify::RecursiveMode::Recursive)?;

    loop {
        match watch_recv.recv() {
            Ok(event) => match event {
                DebouncedEvent::Create(p) | DebouncedEvent::Remove(p) | DebouncedEvent::Write(p) => {
                    let extension = p.extension().unwrap_or_default();
                    if extension != "xml" && extension != "scss" {
                        continue;
                    }

                    let (daemon_resp_sender, mut daemon_resp_response) = tokio::sync::mpsc::unbounded_channel();
                    evt_send.send(app::DaemonCommand::ReloadConfigAndCss(daemon_resp_sender))?;
                    tokio::spawn(async move {
                        match daemon_resp_response.recv().await {
                            Some(app::DaemonResponse::Success(_)) => println!("Reloaded config successfully"),
                            Some(app::DaemonResponse::Failure(e)) => eprintln!("Failed to reload config: {:#?}", e),
                            None => eprintln!("No response to reload configuration-reload request"),
                        }
                    });
                }
                _ => {}
            },
            Err(e) => eprintln!("watch error: {:?}", e),
        }
    }
}
```

elkowar:
another few things:
- please autoformat :<
- Can we be sure that this doesn't break the server termination stuff? All other places that loop in some task are constantly ready for events that shutdown the daemon - which is not the case here. it _seems_ to work - as in, it seems as though the rust runtime just kills the thread - but can we be _sure_ that this is consistent behaviour? some research here would be good, and maybe a quick comment on the function that explains _why_ it's fine (and again, this relates to the switch to `spawn_blocking`, which may impact this, idk)

Having blocking code in an async application is hard and ugly :(

undefinedDarkness:
> All other places that loop in some task are constantly ready for events that shutdown the daemon
how to do this?

elkowar:
> how to do this?
use `loop_select_exiting` from `application_lifecycle` and have fun! that requires the thing to be running in async properly, which is one of the main reasons why we need async here

src/server.rs
```diff
@@ -103,52 +104,42 @@ fn init_async_part(paths: EwwPaths, ui_send: UnboundedSender<app::DaemonCommand>
...
-    let mut buffer = [0; 1024];
-    let mut event_stream = inotify.event_stream(&mut buffer)?;
+async fn run_filewatch<P: AsRef<Path>>(config_dir: P, evt_send: UnboundedSender<app::DaemonCommand>) -> Result<()> {
+
+    use notify::Watcher;
+
+    let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+    let mut watcher: notify::RecommendedWatcher = notify::Watcher::new_immediate(move |res: notify::Result<notify::Event>| match res {
+        Ok(event) => {
+            tx.send(event.paths).unwrap();
```
elkowar, line 115: Not a fan of this unwrap, as that may _actually_ fail if the loop_select_exiting exits

undefinedDarkness: I can't use a ? here, I think, so what do I do?

elkowar:
```
if let Err(err) = tx.send(event.paths) {
  log::warn!("Error forwarding file update event: {:?}", err);
}
```

PR #143, opened 2021-03-27
(Very) Rudimentry gif support

undefinedDarkness:
## Description
If you used an image tag with a gif image it would be empty  ~ this fixes that
by using a gdkpixbufanimation

## Usage
`<image path="image.gif" />`

## Additional Notes
Kazam is garbage.

## Checklist

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,6 +5,8 @@
 #![feature(result_cloned)]
 #![feature(try_blocks)]
 #![feature(nll)]
+#![feature(str_split_once)]
+#![feature(or_patterns)]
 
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -5,6 +5,8 @@
 #![feature(result_cloned)]
 #![feature(try_blocks)]
 #![feature(nll)]
+#![feature(str_split_once)]
+#![feature(or_patterns)]
```
elkowar, line 9: if you update your nightly, these shouldn't be necessary
```suggestion
```

PR #164, opened 2021-05-10
Bumping the gtk version

snakedye:
## Description

Bumping the version of gtk to 3_22 to improve the stability of the Wayland backend.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,8 @@ no-x11-wayland = []
 version = "0.10.0"
 
 [dependencies]
-gtk = { version = "0.9", features = [ "v3_16" ] }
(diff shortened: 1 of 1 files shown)


snakedye:
Eww is neither using gtk 3.24 or gtk-layer-shell 0.6.0 so this message is rather weird.
image

snakedye:
According to [this](https://github.com/wmww/gtk-layer-shell/blob/master/compatibility.md) we should be in the clear but I still get this message :thonk:.
image

buffet (approved):

elkowar:
after some testing, this seems to work. Let's hope that it's not a mistake :P

PR #168, opened 2021-05-13
replaces all daemon prints with logs and enables logging by default

legendofmiracles:
A lot of errors wouldn't be logged with how it is currently, as any logging is disabled without the --debug flag.

With this patch all warnings and errors will be printed, if the --debug flag isn't specified + it also enables the time flags, for easier debugging of script vars and when they execute.

Furthermore, as it's currently there is the issue that eprintln/println and the log crate is used, this tries to minimize the usage of the standard print macros

### DIFF
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -29,14 +29,14 @@ pub(self) fn run_command<T: 'static + std::fmt::Display + Send + Sync>(cmd: &str
             Ok(mut child) => match child.wait_timeout(std::time::Duration::from_millis(200)) {
                 // child timed out
                 Ok(None) => {
-                    eprintln!("WARNING: command {} timed out", &cmd);
+                    log::error!("WARNING: command {} timed out", &cmd);
                     let _ = child.kill();
                     let _ = child.wait();
                 }
-                Err(err) => eprintln!("Failed to execute command {}: {}", cmd, err),
+                Err(err) => log::error!("Failed to execute command {}: {}", cmd, err),
                 Ok(Some(_)) => {}
             },
-            Err(err) => eprintln!("Failed to launch child process: {}", err),
+            Err(err) => log::error!("Failed to launch child process: {}", err),
         }
     });
 }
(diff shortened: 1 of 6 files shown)


elkowar:
I'd argue we should set the log level to `info`, no?

legendofmiracles:
A bit useless stuff in info, the user doesn't really care about the ipc server initializing

And all errors should generally be logged with how it is right now

legendofmiracles:
And eww doesn't really make use of debug or trace, so there would basically be no difference with the flag then

elkowar:
eww logs **a lot** on debug though, specifically with script vars it spams the logs quite hard. Which of the info logs would you deem irrelevant?

legendofmiracles:
Fixed, you're right.

src/widgets/mod.rs
```diff
@@ -106,7 +106,7 @@ fn build_builtin_gtk_widget(
     resolve_widget_attrs(&mut bargs, &gtk_widget);
 
     if !bargs.unhandled_attrs.is_empty() {
-        eprintln!(
+        log::error!(
             "{}WARN: Unknown attribute used in {}: {}",
```
elkowar, line 110: Here the message shouldn't say "warn" anymore

elkowar (changes requested):

shell.nix
```diff
@@ -0,0 +1,22 @@
...
+    atk
+    gdk-pixbuf
+    pango
+    gtk3
+    nixpkgs.latest.rustChannels.nightly.rust
+  ];
+  nativeBuildInputs = [ pkgconfig wrapGAppsHook ];
+
+  RUST_SRC_PATH = "${pkgs.rust.packages.stable.rustPlatform.rustLibSrc}";
+}
```
elkowar, line 22: you won't sneak this in like that :<

legendofmiracles: Oh no! That was an accident, i promise.

I ran git add -A, and since it wasn't added to the .gitignore it got added to the staged files.

elkowar:
Thanks

legendofmiracles:
> I'd argue we should set the log level to info, no?
https://media.discordapp.net/attachments/635625925748457482/844883393316519937/snap_21_05_20_07_15_17.png?width=1128&height=691
now we have this situation, i forgot that they also get printed when running commands.

elkowar:
I'd rather set the command thingies to debug - they're only useful when debugging, after all

elkowar:
U wanna PR?

legendofmiracles:
I got a working solution, i sent it to 'bruh' for testing

elkowar:
Wat

undefinedDarkness:
:sunglasses:

Issue #153, opened 2021-04-10
Set _NET_WM_WINDOW_TYPE and related fields properly


midnightexigent:
I am just getting started with eww. I tried to apply the eww-bar example. 

The problem I am encountring is that the "reserve" attribute doesn't seem to do anything: the bar is displaying over the windows. 

Am I missing something ? 

PS. I am using a tiling window manager "leftwm"

pradyungn:
This is known to have happened in the past, and apparently there's not much you can do atm. Personally I suggest looking into your WM's padding properties for the specified monitor, that way you can just pad it to the space you'd normally reserve.

midnightexigent:
bummer.. it is possible to do that but the padding makes so the bar stays visible even when in full-screen mode

undefinedDarkness:
Yeah reserve is a bit so - so at functioning
Sometimes the window's focusable attribute plays a part tho so try setting that to true

elkowar:
yea the issue is that X's struts stuff is _really_ fucky, and every WM does something different pretty much..... I really don't get how other bars do it, other than just implementing custom code per WM

undefinedDarkness:
> other than just implementing custom code per WM
Sounds like very pain

midnightexigent:
I've done some digging, and I am pretty sure that I found what's wrong. 

TL; DR: `eww` is not asking for the desired behavior from the WM in an "EWMH-compliant way". For a status bar, the `_NET_WM_WINDOW_TYPE` property should be set to `_NET_WM_WINDOW_TYPE_DOCK` (what polybar uses) or `_NET_WM_WINDOW_TYPE_TOOLBAR` whereas `eww` is always setting it to `_NET_WM_WINDOW_TYPE_NORMAL` 

How I found out (long version):

comparing the EWMH properties of an `eww` window and `polybar`, we have two notable differences: `_NET_WM_STATE` and `_NET_WM_WINDOW_TYPE`

polybar: 
```bash
$ xprop -name polybar-mainbar0_DP-4
_NET_WM_PID(CARDINAL) = 24959
_NET_WM_DESKTOP(CARDINAL) = 0
_NET_WM_STATE(ATOM) = _NET_WM_STATE_STICKY, _NET_WM_STATE_ABOVE
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DOCK
WM_CLASS(STRING) = "polybar", "Polybar"
[...]

PR #176, opened 2021-05-17
Set _NET_WM_WINDOW_TYPE and related fields properly

midnightexigent:
## Description

resolving #153 according to [comment](https://github.com/elkowar/eww/issues/153#issuecomment-819362565)

## Additional Notes

This could technically be marked as resolved. But here are some points to consider before merging 
- When `focusable` is `false`, setting `_NET_WM_WINDOW_TYPE` doesn't do anything at all. Should the `focusable` attribute affect the defaults ? 
- `_NET_WM_STATE` also seems important, should there be some way to set it from the top level API ?
- This is exposing some low level stuff to the top-level API and kinda pushes work to the user-side, not to mention that some stuff will break when switching between `wayland` and `x11`. Should there be higher-level abstractions that hide those params ?  

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,8 @@ base64 = "0.13"
 wait-timeout = "0.2"
 
 notify = "5.0.0-pre.7"
+strum_macros = "0.20.1"
+strum = "0.20.0"
 
 [dev-dependencies]
 pretty_assertions = "0.7.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,8 +68,6 @@ base64 = "0.13"
 wait-timeout = "0.2"
 
 notify = "5.0.0-pre.7"
-strum_macros = "0.20.1"
-strum = "0.20.0"
 
 [dev-dependencies]
 pretty_assertions = "0.7.1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -389,6 +389,8 @@ dependencies = [
  "simple-signal",
  "smart-default",
  "structopt",
+ "strum 0.20.0",
+ "strum_macros 0.20.1",
  "sysinfo",
  "tokio",
  "tokio-stream",
@@ -1712,10 +1714,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "57bd81eb48f4c437cadc685403cad539345bf703d78e63707418431cecd4522b"
 
 [[package]]
+name = "strum"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7318c509b5ba57f18533982607f24070a55d353e90d4cae30c467cdb2ad5ac5c"
+
+[[package]]
 name = "strum_macros"
 version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "87c85aa3f8ea653bfd3ddf25f7ee357ee4d204731f6aa9ad04002306f6e2774c"
+dependencies = [
+ "heck",
+ "proc-macro2",
+ "quote 1.0.9",
+ "syn 1.0.72",
+]
+
+[[package]]
+name = "strum_macros"
+version = "0.20.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee8bc6b87a5112aeeab1f4a9f7ab634fe6cbefc4850006df31267f4cfb9e3149"
 dependencies = [
  "heck",
  "proc-macro2",
@@ -1778,8 +1798,8 @@ checksum = "0f3ecc17269a19353b3558b313bba738b25d82993e30d62a18406a24aba4649b"
 dependencies = [
  "heck",
  "pkg-config",
- "strum",
- "strum_macros",
+ "strum 0.18.0",
+ "strum_macros 0.18.0",
  "thiserror",
  "toml",
  "version-compare",
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -220,7 +220,7 @@ For Wayland users the `<reserve/>` block is replaced by the exclusive field in `
 The previous `<window>` block would look like this.
 
 ```xml
-    <window name="main_window" stacking="fg" focusable="false" screen="1" exclusive="true">
+    <window name="main_window" stacking="fg" focusable="false" screen="1" exclusive="true" windowtype="normal">
         <geometry anchor="top left" x="300px" y="50%" width="25%" height="20px"/>
         <widget>
             <main/>
@@ -247,3 +247,10 @@ There are a couple things you can optionally configure on the window itself:
   The details on how it is actually implemented are left to the compositor.
   This option is only valid on Wayland.
   Possible values: `"true"`, `"false"`. Default: `"false"`
+- `windowtype`: (X11 only) Can be used in determining the decoration, stacking position and other behavior of the window.
+  Possible values: 
+    - `"normal"`: indicates that this is a normal, top-level window
+    - `"dock"`: indicates a dock or panel feature
+    - `"toolbar"`: toolbars "torn off" from the main application
+    - `"dialog"`: indicates that this is a dialog window
+Default: `"dock"` if reserve is set, else `"normal"` 
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -253,4 +253,4 @@ There are a couple things you can optionally configure on the window itself:
     - `"dock"`: indicates a dock or panel feature
     - `"toolbar"`: toolbars "torn off" from the main application
     - `"dialog"`: indicates that this is a dialog window
-Default: `"dock"` if reserve is set, else `"normal"` 
+    - Default: `"dock"` if reserve is set, else `"normal"` 
--- a/examples/eww-bar/eww.xml
+++ b/examples/eww-bar/eww.xml
@@ -93,7 +93,7 @@ contain are defined. -->
 
   <windows>
     <!-- These are the windows -->
-    <window name="bar" screen="0">
+    <window name="bar" screen="0" focusable="true" windowtype="dock">
       <geometry x="0%" y="0%" width="100%" height="4%"/> <!--Specifies geometry-->
       <reserve side="top" distance="4%" layer="top"/>
       <widget>
--- a/src/app.rs
+++ b/src/app.rs
@@ -307,7 +307,6 @@ fn initialize_window(
     window_def: config::EwwWindowDefinition,
 ) -> Result<EwwWindow> {
     let actual_window_rect = window_def.geometry.get_window_rectangle(monitor_geometry);
-
     if let Some(window) = display_backend::initialize_window(&window_def, monitor_geometry) {
         window.set_title(&format!("Eww - {}", window_def.name));
         let wm_class_name = format!("eww-{}", window_def.name);
@@ -329,7 +328,7 @@ fn initialize_window(
         gdk_window.set_override_redirect(!window_def.focusable);
 
         #[cfg(feature = "x11")]
-        display_backend::reserve_space_for(&window, monitor_geometry, window_def.struts)?;
+        display_backend::set_xprops(&window, monitor_geometry, &window_def)?;
 
         // this should only be required on x11, as waylands layershell should manage the margins properly anways.
         #[cfg(feature = "x11")]
@@ -344,7 +343,6 @@ fn initialize_window(
     } else {
         Err(anyhow!("monitor {} is unavailable", window_def.screen_number.unwrap()))
     }
-
 }
 
 /// Apply the provided window-positioning rules to the window.
--- a/src/config/window_definition.rs
+++ b/src/config/window_definition.rs
@@ -1,17 +1,31 @@
+use super::*;
 use crate::{ensure_xml_tag_is, value::NumWithUnit, widgets::widget_node};
-use anyhow::*;
 use derive_more::*;
 use serde::{Deserialize, Serialize};
 use smart_default::SmartDefault;
-use std::collections::HashMap;
+use std::{collections::HashMap, str::FromStr};
+use strum_macros::EnumString;
 
-use super::*;
+#[derive(Debug, Clone, PartialEq, EnumString)]
+#[strum(serialize_all = "lowercase")]
+pub enum EwwWindowType {
+    Dock,
+    Dialog,
+    Toolbar,
+    Normal,
+}
+impl Default for EwwWindowType {
+    fn default() -> Self {
+        Self::Normal
+    }
+}
 
 /// Full window-definition containing the fully expanded widget tree.
 /// **Use this** rather than `[RawEwwWindowDefinition]`.
 #[derive(Debug, Clone)]
 pub struct EwwWindowDefinition {
     pub name: WindowName,
+    pub window_type: EwwWindowType,
     pub geometry: EwwWindowGeometry,
     pub stacking: WindowStacking,
     pub screen_number: Option<i32>,
@@ -29,6 +43,7 @@ impl EwwWindowDefinition {
     pub fn generate(defs: &HashMap<String, WidgetDefinition>, window: RawEwwWindowDefinition) -> Result<Self> {
         Ok(EwwWindowDefinition {
             name: window.name,
+            window_type: window.window_type,
             geometry: window.geometry,
             stacking: window.stacking,
             screen_number: window.screen_number,
@@ -51,6 +66,9 @@ pub struct RawEwwWindowDefinition {
     pub screen_number: Option<i32>,
     pub widget: WidgetUse,
     pub focusable: bool,
+
+    #[cfg(feature = "x11")]
+    pub window_type: EwwWindowType,
 
     #[cfg(feature = "x11")]
     pub struts: StrutDefinition,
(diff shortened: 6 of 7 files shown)


Cargo.toml
```diff
@@ -68,6 +68,8 @@ base64 = "0.13"
 wait-timeout = "0.2"
 
 notify = "5.0.0-pre.7"
+strum_macros = "0.20.1"
+strum = "0.20.0"
```
elkowar, line 72: Not a big fan of adding these, writing that tiny bit of `match self` yourself should be fine - no need for additional macro compile time overhead and the inherent issues with generating user-facing config from identifiers in code in a non-obvious way

elkowar:
How would you propose `focusable` to affect the defaults? `focusable=false` seems to be the thing that most people want and use, for several reasons. would this PR change how focusable=true behaves in a way that makes it useful more often?

_NET_WM_STATE does have some things that may be relevant. Do you (or anyone else from #153 ) have any concrete ideas / proposals for a high-level API that exposes the _relevant_ options/combinations to the user here?

In general, I think the goal should be to cover all reasonable usecases in a reasonably easy-to-understand manner. just wrapping X options directly is not going to give us that, I fear - but reaching a full understanding of what all the X options do and how they interact with each other is also rather difficult. A highlevel API wrapper around these would be great.

midnightexigent:
Since `focusable=false` seems to override anything set by `reserve` and `windowtype`, I was thinking of defaulting `focusable` to `true` if  either of those is set

For a high-level wrapper, how about creating a new `enum` (let's call it `WidgetHolder` for now). Each variant represents a high-level 'construct' that represent common ways widgets are grouped together. Each variant exposes attributes that make sense for it. In terms of code it would look something like:

```rust
enum Side {Left, Right, Bottom, Top}
enum WidgetHolder {
    Bar(Side, ...), // a bar like in the example
    Edge{side: Side, struts: StrutDefinition, exclusive: bool, ...}, // something on the edge of the screen(eg. conky-like widget)
    Basic{resizable: bool, movable: bool, titlebar: bool, ...}, // a basic normal window
    Floating{opacity: i32, ...} // a floating window
    ...
    UserDefined(HashMap<String,String>) // user can go crazy, defining their own WidgetHolder by providing a list key-value that will be passed to the backend (tho could be a bad idea to give that much customization to user)
}
impl WidgetHolder {
    #[cfg(feature="x11"))]
    fn apply_for_x11(&self, ...){...}
    #[cfg(feature="wayland"))]
    fn apply_for_wayland(&self, ...){...}
}
```

Each variant knows exactly what x11/wayland properties to set with minimal complexity on the user's-side 

`window` is now comprised of a `WidgetHolder` and attributes that are common to all variants (ie. monitor, geometry, name). It delegates the whole backend part to `WidgetHolder` and only defines generic

elkowar:
Hmmmmm...
I generally don't dislike the idea of wrapping windows up in different window types, but I do see some issues with it:
- We'd need to be _very_ careful to still offer everything that X/wayland allow for via that API. Not a fan of the UserDefined thingy, as that'd make both the code and config rather complex. - but I'd not want to make things impossible for no reason.
- We'd be duplicating a good bit of code
- the API would probably still be rather confusing, especially if you want to do anything that's not directly named


My biggest issue here is that I don't really feel as though I (or any of us) know/understand what exactly X/Wayland offer in terms of options, and how these interact with each other.
My general understanding is that wayland won't be an issue, as reaching the equivalents of X behaviour on wayland shoud just work™️, so we probably only need to worry about X11 for now.

To get some clarity, I'd say a good approach would be to first collect all options that could be provided on X. By that I mean that we should collect every thing that we can set on a window on X, either with GTK or with direct X-calls, and then figure out how the GTK options interact with the X options (specifically, most of them just set X options, so they can be directly related).

Then, once we know everything that _can_ be set on it's own, we can connect the dots and figure out how these options relate to each other. How, for example, override-redirect interacts with _NET_WM_WINDOW_TYPE (if at all), etc etc. That way, we can get a full understanding of the entire API surface we can/should cover, and can understand which options make sense.

My reason for this would be that I'd optimally like to figure out which options are mutually exclusive, or imply each other, and then design an API that allows the user to only set those options in ways that _don't_ conflict with each other (as long as possible). I think designing an API without knowing exactly what we have available in the background doesn't really make sense.

I get that this may be quite a big scope-change for this PR (as it's suddenly about _all_ options rather than just _NET_WM_WINDOW_TYPE), but I feel like just adding support for this one specific option wouldn't really solve the overarching issue of eww's window-configuration being not only confusing but also fragile. A general overhaul is necessary here, and I think we should start it with some more research

elkowar:
If you _wan't_ to continue working on that, I'd more than appreciate it! If you don't, however, as it's a large scope-change/complexity increase, that's of course also fully understandable!

midnightexigent:
> I don't really feel as though I (or any of us) know/understand what exactly X/Wayland offer in terms of options, and how these interact with each other
>  I'd optimally like to figure out which options are mutually exclusive, or imply each other

IDK about how feasible that (understanding interactions between xprops) actually is. [The EWMH spec](https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html) (specifically [this section](https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45850117530832)) has all the needed information about xprops. It is short and sweet and I've already read through it a couple of times. It can be boiled down to 1 sentence: if property A is set, then, the WM should honor it in B way. It doesn't have any notion of mutual exclusivity/properties affecting each other. Trying to logically deduce which options imply each other/are mutually exclusive is quite the leap as that kind of behavior is probably for the WM to decide (eg. a tiling WM might consider 2 options to be mutually exclusive while a regular WM will honor both of them). 

What I am essentially saying is: if (for example) we were to make the assumption that options A and B are mutually exclusive, since the spec doesn't back up that assumption, we'd need to verify on every single WM that the assumption holds. 
That sounds pretty difficult to accomplish 


The general idea behind [the design in my previous comment](https://github.com/elkowar/eww/pull/176#issuecomment-845086846) is to support a specific list of use-cases and for each use-case, define the *canonical* xprops that need to be set. For example, the canonical xprops for the variant `Bar(Side)` would be 

`_NET_WM_WINDOW_TYPE=_NET_WM_WINDOW_TYPE_DOCK`
`_NET_WM_STATE=_NET_WM_STATE_STICKY, _NET_WM_STATE_ABOVE`
`_NET_WM_STRUT` and `_NET_WM_STRUT_PARTIAL` deduced from `Side`  


This obviously assumes that  we are on a WM that *is* EWMH-compliant (which can be checked with `_NET_SUPPORTING_WM_CHECK`)

> A general overhaul is necessary here, and I think we should start it with some more research
> If you wan't to continue working on that, I'd more than appreciate it!

This seems to come up at pretty opportune moment since the eww config is being reworked. Personally idk much about gtk, but I've already read a bunch of documentation regarding xprops. I am available to keep working on this/keep the discussion going.

elkowar:
Specifically seeing _how_ confusing it can be to debug issues in the behaviour of eww related to X and specific window managers (see #178 , for example), I fear that adding additional abstraction without a _lot_ of simplification might not be worth the added complexity in debugging. By hiding more of the interaction with the WM we could make the API more consistent between wayland and X11, and possibly improve the UX in general (at the cost of some flexibility, possibly). _but_ i can see how that would make it even harder to figure out _why_ stuff doesn't work, if it doesn't work.

My big issue with splitting up the configuration into canonically supported variants as you proposed is that I fear it wouldn't add _that_ much value to the user (as in the end, it pretty much just serves as a default, with then some more limited api on top).
Instead, If it comes down to just doing sane defaults and making sure the API is actually understood (which, for example for the _NET_WM_STRUTS stuff, I _really_ can't say I did), I'd propose that it would be a good idea to just have a `window-kind` property (not sure on the name) that sets default options, but then keep all of the fields open all the time. (-> there's just not real gain in limiting them)

I think the main steps in that case would be to
- figure out all the X-api that is actually relevant/useful here, and design a relatively slim abstraction over them
- document these options as good as possible
- see if it's possibly to keep most of them the same between wayland and X, or if it would make more sense to just fully split the configuration into separate config APIs for wayland/X, such that there is no weird layer of pseudo-abstraction in between the two.

midnightexigent:
That sounds good to me

For step 1:

> figure out all the X-api that is actually relevant/useful here

That depends on which use-cases are supported. Are those currently known ? 

> design a relatively slim abstraction over them

I can start looking into draft designs by taking into account what has been mentioned during this discussion. I assume that backwards compatibility isn't a problem, right ?

elkowar:
any updates on this? @midnightexigent

midnightexigent:
Not really.  I haven't had much time to look at this. I'll try to make some next week

elkowar:
This seems pretty good for now! I think having a more high-level abstraction on top of this might be good in the future, but for now, this seems reasonable!
The one thing that I'd like to see here is that we should probably now actually get rid of `focusable` as it is right now.
Specifically, it's currently used to set 3 things:
- override redirect (which should now _generally_ no longer be necessary. This should probably still be possible to set, but under a more exact name like `wm-ignore`.
- To set if the window is keyboard-interactable on wayland. Here we could either rely on the windowtype as provided for X (and thus make it non-X-exclusive), or keep focusable - but restrict it to doing _only_ this.
- Decide the GTK window type. Here, the implications of what Popup vs. toplevel means are somewhat unclear. I'm assuming that _most_ of those are overridden by the other things we set via X directly and via GTK,... but is it _all_ of them? Here, I think the best path of action would be to either
  - figure out _exactly_ what it does, and how it interacts with the other options (honestly,... unrealistic, lol)
  - Set a sane default, and allow this to be set explicitly if so desired. Then, when users face issues, we have another thing to tell them to try out, and can thus get more information on how it behaves. My big issue here is that the additional layer of abstraction that GTK introduces _together with the fact that many if not all WMs behave differently and non-spec-compliant_ makes it incredibly hard to actually know what we need here. That sane default can thus be figured out by checking out other, similar projects using GTK (I'll do this, and report back).

Getting rid of focusable and possibly setting the (gtk) window type to TopLevel everywhere would mean that we'd need to add some more options to "reproduce" what (gtk) window type Popup gives us: Removing borders, for example.

Additionally, it might make sense to use this opportunity to provide the option to explicitly set sticky on a window with a new `sticky` option.

elkowar:
However, writing the last comment, I realized that we're actually not _that_ far away from a higher level API currently.

In general, I can see the following general types of windows that one would want to make with eww, and the following desired behaviour:
- a bar
  - reserves space on a screen edge
  - Sticky (_NET_WM_STATE_STICKY)
  - window type: Dock (_NET_WM_WINDOW_TYPE_DOCK)
  - not focusable (WM does not give it explicit focus. still needs mouse interactability)
  - does not show up in taskbars / related tools (NET_WM_STATE_SKIP_TASKBAR)
- a popup representing a dashboard (i.e. axarvas/adis setup), a popup showing a volume slider, music controls, etc
  - Sticky (_NET_WM_STATE_STICKY)
  - window type: unclear? Dunst uses `_NET_WM_WINDOW_TYPE_NOTIFICATION, _NET_WM_WINDOW_TYPE_UTILITY` as it's two window types. I assume that UTILITY is the "correct" one here, although DOCK may also have the desired result in many window managers. This would need to be tested.
  - not focusable (WM does not give it explicit focus. still needs mouse interactability)
  - does not show up in taskbars / related tools (NET_WM_STATE_SKIP_TASKBAR)
  - always on top (_NET_WM_STATE_ABOVE) (EWMH does not actually specify how a WM should decide on it's stacking order, so this is inherently unspecified behaviour - although ABOVE is definitely the accepted convention here)
- A widget shown in the background, showcasing system stats, time, etc (conky, rainmeter)
  - not focusable  (WM does not give it explicit focus. still needs mouse interactability)
  - sticky
  - window type: UTILITY (probably, also not fully clear here tho. might also be DESKTOP?)
  - does not show up in taskbars / related tools  (NET_WM_STATE_SKIP_TASKBAR)
  - always behind (_NET_WM_STATE_BELOW). There have been issues that reported that stacking `bg` actually places eww _behind the wallpaper_,.... some WMs seem to _want_ to cause me pain. This needs further investigation, but not necessarily in the scope of this.
- a popup window for dialogs, menus, etc
  - Not sticky
  - _may_ show up in taskbars? (what icon would be set here?)
  - focusable and fully interactable
  - window type: _NET_WM_WINDOW_TYPE_DIALOG or _NET_WM_WINDOW_TYPE_NORMAL
  - acts as a normal window. no special always-in-front or always-behind stuff here.
  - _may_ still not be desired to get tiled by window managers,...

elkowar:
Either way, coming back to this PR concretely: depending on how much time and motivation you have right now, I'd say we could either add (some of) the stuff that I mentioned in my first comment, or leave it as it is. The current implementation is generally ready-to-merge in my opinion, so splitting off the next steps into separate PRs might be sensible. If however you have tons of time and energy right now, I'd obviously be more than greatful for addition of those things in the scope of this PR.

So, TL;DR: If you feel as though you're done with this, I'd be happy to merge! If you wanna continue work on this, then that'd also be great!

midnightexigent:
Hi again,


First of all, I do agree that the goal of the initial PR has been implemented, so fine by me if you want to merge as is. 

Now to the good stuff. So I got around to start looking into the `EwwWindowDefinition` overhaul discussed earlier. Very early WIP for now tho, but here's a general idea

1. Seperate backend-specific configs into 2 modules

```rust 

#[cfg(feature = "x11")]
mod x11 {
    struct X11 {/*...*/}

    struct StrutDefinition{/*...*/}
    enum WindowType {/*...*/}

    /*...*/
}

#[cfg(feature = "wayland")]
mod wayland {
    struct Wayland {/*...*/}
    
    /*...*/
}

#[cfg(feature = "x11")]
type Backend = x11::X11;

#[cfg(feature = "wayland")]
type Backend = wayland::Wayland;

```

2. Have the `Preset` enum, which designed primarly to give a way for the user to generate sane predefined `EwwWindowDefinition`s with minimal effort

```rust
pub enum Preset {
    Bar{ /*...*/ },
    Floating{ /*...*/ },
    /*...*/
}

// generate EwwWindowDefinition attributes from the preset

impl TryFrom<&Preset> for Backend {/*...*/}

impl TryFrom <&Preset> for EwwWindowGeometry {/*...*/}

/*...*/
```

3. With the previous 2 points taken into account `RawEwwWindowDefinition` looks like this

```rust

// in the impl, for each field, if it is not provided, `preset` is the fallback (if applicable) 
struct RawEwwWindowDefinition {
      widget: WidgetUse, //preset can't help us here
      preset: Option<Preset>, 
      geometry: Option<EwwWindowGeometry>, // or preset.try_into()? or default()

      backend: Option<Backend>, // or preset.try_into()? or default()

      /*...*/
}

```


That being the general idea and what we discussed, I am now looking into making this more sophisticated using [derive_builder](https://docs.rs/derive_builder/0.10.2/derive_builder/) and [merge](https://docs.rs/merge/0.1.0/merge/) to allow for fine-grained overriding. 

I think this is pretty much necessary since with the above layout, the user has to either use the value provided by the preset OR to spell out every single field - a bit cumbersome. One way to achieve fine-grained overriding is to apply `#[derive(Builder)]; #[builder(derive(Merge))]`  on the types used in `RawEwwWindowDefinition`. Essentially, this allows any part of the WindowDef to be deep-merged from multiple sources (preset, inference from another attribute or user-override) . This changes a couple of things, here's a concrete example of how a `Backend` would be generated

```rust 
struct RawEwwWindowDefinition {
    preset: Option<Preset>,
    backend_builder: Option<BackendBuilder>,
    /*...*/
}


let mut backend_builder = raw_eww_window_def.backend_builder.unwrap_or_default();

// infer some BackendBuilder attributes from WindowStacking and EwwWindowGeometry ? 
backend_builder.merge(BackendBuilder::try_from((raw_eww_window_def.stacking, raw_eww_window_def.geometry))?) 

// apply the preset if it is defined
raw_eww_window_def.preset.and_then(|v|backend_builder.merge(v))

backend_builder.build()
```

To summarize, here are the main advantages of this design:
- User has the option to use predefined setups and override them 
- Infer/detect nonsensical configs in a structured/modular way by taking advantage of `TryFrom<T> for *`

Before I actually go on to fully implementing this, can I have your thoughts ? 

Some further thoughts/questions

- How useful is gtk actually for this project ? Eww does some pretty low-level stuff, would it make sense to get rid of the middle man (gtk) ? Especially, if the design I am suggesting is used, it is possible let `preset` or the user do the heavy lifting when it comes to interacting with the display backend
- With the whole config rework, I was wondering if `serde` is gonna become an option ?

midnightexigent:
also, maybe we should move the discussion to a new PR ?

VuiMuich:
This looks fantastic to me.
I am also in the favor of dropping GTK, as it appears to come with quite a bunch of restrictions and eww kind of feels to be on its way to be it's own toolkit, especially with the upcoming s-expressions.

elkowar:
Let's maybe actually move this into a discussion, as this is starting to turn into a larger refactor!

w.r.t. getting rid of gtk: the more I think about it the more I want to do it, but,.... well there are a couple issues.
GTK _does_ provide some ginormeous things: 
- incredibly sophisticated layouting (while somewhat anoying at times, it's very powerful)
- a CSS implementation (this is the big one. Implementing a complete layouting + styling engine that implements something even remotely as powerful as even GTKs limited subset of CSS is.... a incredibly difficult task. While there are some other rust gui frameworks, these tend to not have anything similar to the CSS themability - which would mean a _lot_ of additional things to implement
- cross plattform usability. While icky at times, gtk does work everywhere (mostly). Macos compatibility _should_ be doable, it's just pain for me to work on without a mac. I don't know how flawless this would work everywhere else
- some of GTKs features are very complex to replicate again - such as animations, for example.

A move away from GTK may at some point happen, but most likely not in the current state of the ecosystem, and most likely not with the current amount of changes already going on in eww. Implementing a GUI framework from scratch or even building on top of another one is,.... a ginormeous task.

elkowar:
@midnightexigent I _would_ merge this for now, but currently the wayland build is failing, so that needs fixing ^^ https://github.com/elkowar/eww/pull/176/checks?check_run_id=3075492279#step:7:51

elkowar:
I opened #216, let's move the discussion there

elkowar:
Thanks a ton!

Issue #175, opened 2021-05-17
[bug] Multiple screen configuration breaks stuff


manderio:
When I have multiple monitors configured under screen using ewwbar, it gets confused during reload when a monitor is missing, resulting in this:
![image](https://user-images.githubusercontent.com/59826300/118520239-32035f00-b73a-11eb-848a-6ce742765e1a.png)

Got some eww debug output attached [here](https://github.com/elkowar/eww/files/6495248/debug.txt)

as for configuration, got it right here: [ewwxml](https://github.com/elkowar/eww/files/6495259/ewwconf.txt)
scss shouldn't matter but got that too just in case: [ewwscss](https://github.com/elkowar/eww/files/6495266/ewwconf.txt)

Additional notes, I am sitting on arch on Wayland, if you want anything about my system, please do ask me anything. Also, please have a nice day and a [hug!](https://tenor.com/HZmw.gif)

manderio:
Note, I defined bar1 twice, which leads to the behavior. However, would be nice if I got a warning about this rather than just strange behavior, will therefore keep issue open :D

elkowar:
TODO:
handle monitor selection properly.
- in display_backend under wayland, `get_monitor` may return None when the monitor ID is invalid. in that case, the window should probably not be opened at all -> make monitor-id required.

- to get monitor-geometry, the code defaults to getting the default screen if none is given, but does not change that in the window_definition. Thus, the display_backend is given out-of-sync data, which should be fixed.

manderio:
Hello, more additional notes, this breaks when using open-many too:
![image](https://user-images.githubusercontent.com/59826300/118781170-0ba20880-b88d-11eb-8cdd-8cdeea7cb8a8.png)

Edited the configs above to remove the mistakes mentioned, part looks like this now:
```xml
	    <!-- bar -->
	    <!-- main screen -->
	    <window name="bar0" stacking="bt" screen="0" exclusive="true">
		<geometry anchor="top center" width="100%" height="25px"/>
		<widget>
		    <bar0/>
		</widget>
	    </window>

	    <!-- for external monitor 1 -->
	    <window name="bar1" stacking="bt" screen="1" exclusive="true">
		<geometry anchor="top center" width="100%" height="25px"/>
		<widget>
		    <bar1/>
		</widget>
	    </window>

		<!-- for external monitor 2 -->
		<window name="bar2" stacking="bt" screen="2" exclusive="true">
[...]

PR #180, opened 2021-05-22
Fix for #175.

snakedye:

## Description
Widgets on non existing monitors do not spawn. 

## Additional Notes

Currently this is only effective for Wayland.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -308,41 +308,43 @@ fn initialize_window(
 ) -> Result<EwwWindow> {
     let actual_window_rect = window_def.geometry.get_window_rectangle(monitor_geometry);
 
-    let window = display_backend::initialize_window(&window_def, monitor_geometry);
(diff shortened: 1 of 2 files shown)


snakedye:
@manderio testing, testing, 1, 2, 3

manderio:
it works fine for me too, also really like the improved error messages, they look really neat! 

[Thanks for being epic and fixing this!](https://tenor.com/7Rmn.gif)

buffet (approved):
Seems good, no way to explicitly do pass NULL to the compositor though, in case that's what the user wants.

Not sure how feasible that is w.r.t. X though.

elkowar:
^ this - having a way to state "on current screen" would be nice, both on X and on wayland. 

Also build fails, check out the error and fix 👀

snakedye:
The default behavior when no screen is defined is current screen with the Wayland backend.

elkowar:
Thanks!

PR #181, opened 2021-05-23
logging: sensible defaults

legendofmiracles:
## Description
Makes the default logging different dependent on what type of command is running. 
It also delays the printing of the starting eww daemon message, until after the dispatch, so that we can stay with the info filter.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -35,14 +35,6 @@ pub mod widgets;
 fn main() {
     let opts: opts::Opt = opts::Opt::from_env();
 
-    let log_level_filter = if opts.log_debug { log::LevelFilter::Debug } else { log::LevelFilter::Info };
-    if std::env::var("RUST_LOG").is_ok() {
-        println!("hey");
-        pretty_env_logger::init_timed();
-    } else {
-        pretty_env_logger::formatted_timed_builder().filter(Some("eww"), log_level_filter).init();
-    }
-
     let result: Result<()> = try {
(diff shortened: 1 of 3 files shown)


elkowar:
Not really a big fan of this - don't see much benefit in doing it,... or am I missing something?

legendofmiracles:
This will stop this behavior: https://media.discordapp.net/attachments/635625925748457482/844883393316519937/snap_21_05_20_07_15_17.png?width=1128&height=691

elkowar:
shouldn't those INFOs just really be DEBUG? No reason for anything super complex, just change the `log::info` to `log::debug`, no?

legendofmiracles:
Whoo

legendofmiracles:
oh fuck

legendofmiracles:
uhh, if you ignore that @elkowar sorry for not responding here.

The reason why it should be info, and should be done in the way I did it, is so that when running eww commands stuff doesn't get logged in your terminal, but when you run `eww logs` you will see it. 

I've noticed that sometimes eww doesn't log important things when running in non-debug mode, so this would fix that.

If you still want this PR, then I'll open a new one since i cherry-picked my commit onto a new branch, and don't think that there's a way to change the branch of a already opened PR.

elkowar:
Yes, but thats my point - shouldn't we just change this log level of those said important things, rather than changing which logs we include in weird ways?

PR #185, opened 2021-06-03
Added another example setup in README.md

adi1090x:
### Description
Just added another example section of my setup.

### Usage
Help people to understand possibilities.

### Showcase

Added screenshot

### Additional Notes

Let me know if i should add this config in example directory.

**PS** : we really need `onscrollup` `onscrolldown` mouse events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,9 @@ Please give your opinions in the [discussion post](https://github.com/elkowar/ew
 * [Setup by Axarva](https://github.com/Axarva/dotfiles-2.0)
 ![Axarva-rice](https://raw.githubusercontent.com/Axarva/dotfiles-2.0/main/screenshots/center.png)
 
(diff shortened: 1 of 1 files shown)


elkowar:
Hey, thanks for the PR! Given the relative complexity of your setup, I'd say it's best to keep it as a link to your repo rather than duplicating the code into the examples here - it might otherwise be a bit daunting as a reference for new people. 
I'll merge this!

*PS*: widgets do have the `onscroll` property, which takes a command that will get executed when scrolling. In that command you can use the `{}` placeholder, simmilar to the onchange command on sliders, which will get replaced with either "up" or "down". I just realized that that placeholder was not documented, that should be fixed now! 
the onscroll miiiiiight not quite work on all widgets, as GTK is a bit,.. fucky in that regard, but it'll work everywhere where gtk let's me do it :p

PR #209, opened 2021-07-12
Added angle property for label widget

RedstoneWizard22:
## Description

Added the ability to set the angle of the label widget. Useful for sidebars :D

## Usage

```xml
<label angle="90" text="Awesome!"/>
```
angle is a double, and will default to 0 (horizontal) if omitted

### Showcase

image

### DIFF
--- a/src/widgets/widget_definitions.rs
+++ b/src/widgets/widget_definitions.rs
@@ -502,7 +502,7 @@ fn build_gtk_label(bargs: &mut BuilderArgs) -> Result<gtk::Label> {
         prop(wrap: as_bool) {
             gtk_widget.set_line_wrap(wrap)
         },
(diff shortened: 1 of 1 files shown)


src/widgets/widget_definitions.rs
```diff
@@ -501,6 +501,10 @@ fn build_gtk_label(bargs: &mut BuilderArgs) -> Result<gtk::Label> {
         // @prop wrap - Wrap the text. This mainly makes sense if you set the width of this widget.
         prop(wrap: as_bool) {
             gtk_widget.set_line_wrap(wrap)
+        },
+        // @prop angle - the angle of rotation for the label
```
elkowar, line 505:
```suggestion
        // @prop angle - the angle of rotation for the label (between 0 - 360)
```

PR #217, opened 2021-07-15
Initial X Window configuration cleanup

elkowar:
This PR lays some groundwork for #216 by adding more flexibility for telling X how it should treat windows.
To implement that, it separates options more cleanly between the wayland and X backend, removes `focusable` as an option on X and replaces it with the `wm-ignore` flag, which now specifically just set's override redirect.
Additionally, it adds some other new options:
- sticky
- resizable
as well as adding the "utility" window type.

This PR is a breaking change in some configurations, as it will change how window positioning behaves depending on how the configuration looks currently.
Some smart defaults are implemented to avoid issues in most cases: wm-ignore will default to true (making eww behave similarly to how it did before), except if either reserve or windowtype are set. window-type defaults to dock if reserve is set.

# migration
setups that previously relied on focusable="true" can now be implemented using `wm-ignore="false"`. If the setup is a bar or similar, `windowtype="dock"` should be set, to properly apply reserve without making eww appear in application selectors or other places.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "e8fd72866655d1904d6b0997d0b07ba561047d070fbe29de039031c641b61217"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


undefinedDarkness:
Pog!

elkowar:
This is integrated into master since the config rework, and can thus be closed

PR #218, opened 2021-07-16
inotify scheduler for script vars

legendofmiracles:
## Description

Adds another scheduler for script-vars, which will execute the command whenever a file/directory changes.

## Usage

e.g. 
```xml
<script-var name="test" files="[/home/user/test,/sys/class/power_supply/]/>
```

## Additional Notes

See the docs I added.

### DIFF
--- a/docs/src/configuration.md
+++ b/docs/src/configuration.md
@@ -76,7 +76,7 @@ Useful for creating volume sliders or anything similar.
 Example:
 ```xml
 <variables>
-    <script-var name="date" interval="5s">
+    <script-var name="date" interval="5s" files=[/home/user/time,/root/time]>
         date +%H:%M
     </script-var>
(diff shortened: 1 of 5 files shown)


elkowar:
Do you have a concrete usecase example for this? In it's current state, this seems,.. rather confusing (both in terms of code, and for the user) - where would a simple script-var (if necessary, using sth like `entr`) not suffice?

Animeshz:
Can't we just use a listening variable for this? We can easily spawn bg child processes (with `&`) monitored by linux itself. Making sleep/loop is easy either.

undefinedDarkness:
Cant you just use entr or smth

legendofmiracles:
Fine, I'm too lazy to resolve the merge conflicts anyways..

PR #226, opened 2021-08-07
Adds a notice to check the logs when the eww daemon isn't running

legendofmiracles:
## Description

Maybe this will stop the huge amount of issues that say their eww daemon doesn't start

## Usage
It just works:tm:

### Showcase

image


elkowar:
Now that the eww daemon autostarts and logs are printed out better, is this still necessary?

PR #239, opened 2021-08-21
Better EWW_RAM / Docs update

druskus20:
## Description

This PR:
- adds information to the magic variable EWW_RAM. 
- renames EWW_CPU_USAGE to EWW_CPU given that it also has frequency
- it also renames some of the functions related to magic variables (system_stats.rs) for the sake of consistency.  
- updates the docs regarding magic variables briefly

## Additional Notes

Using EWW_CPU_USAGE is still okay, but I've added a deprecation warning. 

## Checklist

### DIFF
--- a/crates/eww/src/config/system_stats.rs
+++ b/crates/eww/src/config/system_stats.rs
@@ -40,7 +40,7 @@ pub fn get_ram() -> String {
     )
 }
 
-pub fn get_core_temperatures() -> String {
+pub fn get_temperatures() -> String {
     let mut c = SYSTEM.lock().unwrap();
     c.refresh_components_list();
     c.refresh_components();
--- a/crates/eww/src/config/system_stats.rs
+++ b/crates/eww/src/config/system_stats.rs
@@ -7,7 +7,7 @@ use sysinfo::{ComponentExt, DiskExt, NetworkExt, NetworksExt, ProcessorExt, Syst
 
 static SYSTEM: Lazy<Mutex<System>> = Lazy::new(|| Mutex::new(System::new()));
 
-pub fn disk() -> String {
+pub fn get_disks() -> String {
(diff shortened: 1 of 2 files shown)


crates/eww/src/config/system_stats.rs
```diff
@@ -7,32 +7,49 @@ use sysinfo::{ComponentExt, DiskExt, NetworkExt, NetworksExt, ProcessorExt, Syst
...
+    let used_memory = c.get_used_memory();
+    format!(
+        r#"{{"total_mem": {}, "free_mem": {}, "total_swap": {}, "free_swap": {}, "available_mem": {}, "used_mem": {}, "used_mem_perc": {}}}"#,
+        total_memory,
+        c.get_free_memory(),
+        c.get_total_swap(),
+        c.get_free_swap(),
+        c.get_available_memory(),
+        used_memory,
+        used_memory / total_memory,
```
elkowar, line 48: This isn't how one calculates percentages,..... this needs *100 somewhere, and then rounding _after_ that, not before

crates/eww/src/config/system_stats.rs
```diff
@@ -7,32 +7,49 @@ use sysinfo::{ComponentExt, DiskExt, NetworkExt, NetworksExt, ProcessorExt, Syst
...
-                r#""{}": {{"name": {:?}, "total": {}, "free": {}}}"#,
-                c.get_mount_point().display(),
-                c.get_name(),
-                c.get_total_space(),
-                c.get_available_space(),
-            ))
+            .map(|c| {
+                let total_space = c.get_total_space();
+                let available_space = c.get_available_space();
+                format!(
```
elkowar, line 21: Given that we're providing percentages for ram now, would make sense to also have a `used_perc` here

elkowar (changes requested):

PR #258, opened 2021-09-02
I3 dock fix

### LINKED ISSUE #197: [BUG] Reserve doesn't work in i3
Reserve doesn't work in i3

Config:
```xml
<eww>
    <includes>
        <!-- Put your <file>'s in here -->
    </includes>

    <definitions>
        <def name="icon_button">
            <button>
                <label text="{{desktop_number}}"/>
            </button>
        </def>
        <def name="desktop_switch">
            <button>
                <label text="{{desktop_number}}"/>
            </button>
        </def>
        <def name="bar">
            <box class="main_box_bar_window">
                <box class="right" halign="start">
                    <button class="search_button">
                        <image path="./search_icon_2.png" height="22"/>
                    </button>
                    <desktop_switch desktop_number="1"/>
                </box>
                <box class="middle">
                    <label text="Todo: jdeokfdeofk"/>
                </box>
                <box class="end" halign="end">
                    <button>
                    {{date}}
                    </button>
                </box>

            </box>
        </def>
    </definitions>

    <variables>
        <script-var name="date" interval="5s">
        date "+%a %b %d   %-I:%M %p"
        </script-var>
    </variables>

    <windows>
        <window name="bar_window" stacking="fg" focusable="false">
            <geometry anchor="top center" x="0" y="0" width="100%" height="40"/>
            <reserve side="top" distance="40px"/>
            <widget>
                <revealer transition="slidedown" duration="600ms" reveal="true">
                    <bar/>
                </revealer>
            </widget>
        </window>
    </windows>
</eww>
```

### LINKED ISSUE #248: [BUG] Large CPU utilization when creating a docked widget
## Describe the bug
This is a weird one. Likely a user error or related to some parts of my setup, but I'd appreciate a pointer where to start looking.

## How to reproduce
I doubt this is easily reproducible, but here's my setup:

* Linux 5.13.12-arch1-1 x86_64 GNU/Linux
* i3 version 4.19.2 (2021-02-27)
* eww at 70285e0e
* Screen 0: minimum 320 x 200, current 3840 x 2160, maximum 16384 x 16384
eDP-1 connected primary 3840x2160+0+0 (normal left inverted right x axis y axis) 346mm x 194mm

I've got a simple configuration to create a bar:
```
(defwindow bar
           :monitor 0
           :geometry (geometry :x "0%"
                               :y "0%"
                               :width "100%"
                               :height "30px"
                               :anchor "bottom center")
           :stacking "fg"
           :reserve (struts :distance "40px" :side "bottom")
           :wm-ignore false
  "Hello, world!")
```

The first time I spawn the bar with `eww open bar` after starting the daemon, the bar is not properly docked to the bottom part of the screen, the window manager seems to not reserve the space. (I realize there is #197, so I won't dwell on that longer than necessary)
If I close and open the bar again, the window manager does reserve the space for it, but also the eww daemon starts utilizing 70% of one of my cores, with Xorg taking another one at roughly the same utilization. Very chatty those two apparently.

I put some debug logs in to investigate where the daemon uses all those cycles and ended up here:
```diff
diff --git a/crates/eww/src/app.rs b/crates/eww/src/app.rs
index 1e14c6d..918efff 100644
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -362,9 +362,16 @@ fn apply_window_position(
     monitor_geometry: gdk::Rectangle,
     window: &gtk::Window,
 ) -> Result<()> {
+
     let gdk_window = window.get_window().context("Failed to get gdk window from gtk window")?;
     window_geometry.size = Coords::from_pixels(window.get_size());
     let actual_window_rect = get_window_rectangle(window_geometry, monitor_geometry);
+
+    //let origin = gdk_window.get_origin();
+    //let size = gdk_window.get_geometry();
+    //log::debug!("Moving window to new position: {:?} ({:?}, {:?})", window_geometry, origin, size);
+    log::debug!("Moving window to new position: {:?}", window_geometry);
+
     gdk_window.move_(actual_window_rect.x, actual_window_rect.y);
     Ok(())
 }
```

This produces log output akin to this:
```
 2021-08-30T14:36:34.744Z DEBUG eww::app        > Moving window to new position: WindowGeometry { anchor_point: AnchorPoint { x: CENTER, y: END }, offset: CoordsWithUnits(0, 0), size: CoordsWithUnits(3840, 42) }
 2021-08-30T14:36:34.744Z DEBUG eww::app        > Moving window to new position: WindowGeometry { anchor_point: AnchorPoint { x: CENTER, y: END }, offset: CoordsWithUnits(0, 0), size: CoordsWithUnits(3840, 42) }
 2021-08-30T14:36:34.744Z DEBUG eww::app        > Moving window to new position: WindowGeometry { anchor_point: AnchorPoint { x: CENTER, y: END }, offset: CoordsWithUnits(0, 0), size: CoordsWithUnits(3840, 42) }
 2021-08-30T14:36:34.744Z DEBUG eww::app        > Moving window to new position: WindowGeometry { anchor_point: AnchorPoint { x: CENTER, y: END }, offset: CoordsWithUnits(0, 0), size: CoordsWithUnits(3840, 42) }
 2021-08-30T14:36:34.744Z DEBUG eww::app        > Moving window to new position: WindowGeometry { anchor_point: AnchorPoint { x: CENTER, y: END }, offset: CoordsWithUnits(0, 0), size: CoordsWithUnits(3840, 42) }
 2021-08-30T14:36:34.744Z DEBUG eww::app        > Moving window to new position: WindowGeometry { anchor_point: AnchorPoint { x: CENTER, y: END }, offset: CoordsWithUnits(0, 0), size: CoordsWithUnits(3840, 42) }
 2021-08-30T14:36:34.744Z DEBUG eww::app        > Moving window to new position: WindowGeometry { anchor_point: AnchorPoint { x: CENTER, y: END }, offset: CoordsWithUnits(0, 0), size: CoordsWithUnits(3840, 42) }
```
on repeat.

Interestingly, using the longer log form commented out above _"fixes"_ the problem in that it prevents the above loop from happening, but also again makes the window manager not reserve space for the bar. Not really sure what to make of that.

From what I've read of other people's problems, i3 seems to be _peculiar_ when handling docked windows. My best guess is that i3 renders the bar slightly different to how eww is setting it's geometry, causing a new update message for the window to be fired, which in turn causes eww to again update it's position - and repeat?

## Expected behaviour
Not using 2 cores to render a status bar :sweat_smile: 

## Additional context
Strace of the daemon process:
```
recvmsg(3, {msg_name=NULL, msg_namelen=0, msg_iov=[{iov_base="\226\0M2\20\0\0\1\20\0\0\1\0\0\0\0\0\0F\10\0\17*\0\0\0\0\0\0\0\0\0"..., iov_len=4096}], msg_iovlen=1, msg_controllen=0, msg_flags=0}, 0) = 64
recvmsg(3, {msg_namelen=0}, 0)          = -1 EAGAIN (Resource temporarily unavailable)
clock_gettime(CLOCK_REALTIME, {tv_sec=1630335865, tv_nsec=249624151}) = 0
write(2, " 2021-08-30T15:04:25.249Z \33[0m\33["..., 237) = 237
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=13491597}) = 0
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=13563514}) = 0
poll([{fd=3, events=POLLIN}, {fd=4, events=POLLIN}, {fd=6, events=POLLIN}], 3, 0) = 0 (Timeout)
clock_gettime(CLOCK_REALTIME, {tv_sec=1630335865, tv_nsec=250043943}) = 0
write(2, " 2021-08-30T15:04:25.250Z \33[0m\33["..., 237) = 237
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=13908514}) = 0
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=13979389}) = 0
poll([{fd=3, events=POLLIN|POLLOUT}], 1, -1) = 1 ([{fd=3, revents=POLLOUT}])
writev(3, [{iov_base="\f;\5\0\20\0\0\1\3\0\0\0\0\0\0\0F\10\0\0\f\0\5\0\20\0\0\1\3\0\0\0"..., iov_len=40}, {iov_base=NULL, iov_len=0}, {iov_base="", iov_len=0}], 3) = 40
recvmsg(3, {msg_namelen=0}, 0)          = -1 EAGAIN (Resource temporarily unavailable)
poll([{fd=3, events=POLLIN}, {fd=4, events=POLLIN}, {fd=6, events=POLLIN}], 3, -1) = 1 ([{fd=3, revents=POLLIN}])
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=14365181}) = 0
recvmsg(3, {msg_name=NULL, msg_namelen=0, msg_iov=[{iov_base="\226\0O2\20\0\0\1\20\0\0\1\0\0\0\0\0\0F\10\0\17*\0\0\0\0\0\0\0\0\0"..., iov_len=4096}], msg_iovlen=1, msg_controllen=0, msg_flags=0}, 0) = 64
recvmsg(3, {msg_namelen=0}, 0)          = -1 EAGAIN (Resource temporarily unavailable)
clock_gettime(CLOCK_REALTIME, {tv_sec=1630335865, tv_nsec=250919110}) = 0
write(2, " 2021-08-30T15:04:25.250Z \33[0m\33["..., 237) = 237
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=14802014}) = 0
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=14873306}) = 0
poll([{fd=3, events=POLLIN}, {fd=4, events=POLLIN}, {fd=6, events=POLLIN}], 3, 0) = 0 (Timeout)
clock_gettime(CLOCK_REALTIME, {tv_sec=1630335865, tv_nsec=251336068}) = 0
write(2, " 2021-08-30T15:04:25.251Z \33[0m\33["..., 237) = 237
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=15196681}) = 0
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=15267931}) = 0
poll([{fd=3, events=POLLIN|POLLOUT}], 1, -1) = 1 ([{fd=3, revents=POLLOUT}])
writev(3, [{iov_base="\f;\5\0\20\0\0\1\3\0\0\0\0\0\0\0F\10\0\0\f\0\5\0\20\0\0\1\3\0\0\0"..., iov_len=40}, {iov_base=NULL, iov_len=0}, {iov_base="", iov_len=0}], 3) = 40
recvmsg(3, {msg_namelen=0}, 0)          = -1 EAGAIN (Resource temporarily unavailable)
poll([{fd=3, events=POLLIN}, {fd=4, events=POLLIN}, {fd=6, events=POLLIN}], 3, -1) = 1 ([{fd=3, revents=POLLIN}])
clock_gettime(CLOCK_MONOTONIC, {tv_sec=8376, tv_nsec=15641389}) = 0
```

RianFuro:
## Description

Show window only after settings x-props. According to the EWMH spec:
> _NET_WM_WINDOW_TYPE
> This SHOULD be set by the Client before mapping to a list of atoms
> indicating the functional type of the window.

The callback to the configure event handle now checks the window's
current geometry to not move the window if it's already where it's
supposed to be. This prevents the eww daemon from endlessly chatting
with X11 in some instances.

## Additional Notes

Closes #248
Closes #197

## Checklist

### DIFF
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -345,7 +345,7 @@ fn initialize_window(
 
     window.add(&root_widget);
 
-    window.show_all();
+    window.realize();
 
     #[cfg(feature = "x11")]
     {
@@ -358,6 +358,9 @@ fn initialize_window(
         }
         display_backend::set_xprops(&window, monitor_geometry, &window_def)?;
     }
+
+    window.show_all();
+
     Ok(EwwWindow { name: window_def.name.clone(), definition: window_def, gtk_window: window })
 }
 
@@ -371,7 +374,13 @@ fn apply_window_position(
     let gdk_window = window.get_window().context("Failed to get gdk window from gtk window")?;
     window_geometry.size = Coords::from_pixels(window.get_size());
     let actual_window_rect = get_window_rectangle(window_geometry, monitor_geometry);
-    gdk_window.move_(actual_window_rect.x, actual_window_rect.y);
+
+    let gdk_origin = gdk_window.get_origin();
+
+    if actual_window_rect.x != gdk_origin.1 || actual_window_rect.y != gdk_origin.2 {
+        gdk_window.move_(actual_window_rect.x, actual_window_rect.y);
+    }
+
     Ok(())
 }
 
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -274,7 +274,6 @@ impl App {
         // refresh script-var poll stuff
         self.script_var_handler.stop_all();
 
-
         log::trace!("loading config: {:#?}", config);
 
         self.eww_config = config;
--- a/crates/eww/src/opts.rs
+++ b/crates/eww/src/opts.rs
@@ -173,7 +173,7 @@ fn parse_var_update_arg(s: &str) -> Result<(VarName, DynVal)> {
 
 impl ActionWithServer {
     pub fn can_start_daemon(&self) -> bool {
-        matches!(self, ActionWithServer::OpenWindow {..} | ActionWithServer::OpenMany { .. })
+        matches!(self, ActionWithServer::OpenWindow { .. } | ActionWithServer::OpenMany { .. })
     }
 
     pub fn into_daemon_command(self) -> (app::DaemonCommand, Option<daemon_response::DaemonResponseReceiver>) {
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -8,7 +8,7 @@ use gdk::WindowExt;
 use glib;
 use gtk::{self, prelude::*, ImageExt};
 use itertools::Itertools;
-use std::{cell::RefCell, collections::HashMap, rc::Rc, time::Duration, cmp::Ordering};
+use std::{cell::RefCell, cmp::Ordering, collections::HashMap, rc::Rc, time::Duration};
 use yuck::{
     config::validate::ValidationError,
     error::{AstError, AstResult, AstResultExt},
@@ -41,7 +41,11 @@ pub(super) fn widget_to_gtk_widget(bargs: &mut BuilderArgs) -> Result<gtk::Widge
         "revealer" => build_gtk_revealer(bargs)?.upcast(),
         "if-else" => build_if_else(bargs)?.upcast(),
         _ => {
-            return Err(AstError::ValidationError(ValidationError::UnknownWidget(bargs.widget.name_span, bargs.widget.name.to_string())).into())
+            return Err(AstError::ValidationError(ValidationError::UnknownWidget(
+                bargs.widget.name_span,
+                bargs.widget.name.to_string(),
+            ))
+            .into())
         }
     };
     Ok(gtk_widget)
@@ -517,11 +521,18 @@ fn build_center_box(bargs: &mut BuilderArgs) -> Result<gtk::Box> {
             // we know that there is more than three children, so unwrapping on first and left here is fine.
             let first_span = additional_children.first().unwrap().span();
             let last_span = additional_children.last().unwrap().span();
-            Err(DiagError::new(gen_diagnostic!("centerbox must contain exactly 3 elements, but got more", first_span.to(last_span))).into())
+            Err(DiagError::new(gen_diagnostic!(
+                "centerbox must contain exactly 3 elements, but got more",
+                first_span.to(last_span)
+            ))
+            .into())
         }
         Ordering::Equal => {
-            let mut children =
-            bargs.widget.children.iter().map(|child| child.render(bargs.eww_state, bargs.window_name, bargs.widget_definitions));
+            let mut children = bargs
+                .widget
+                .children
+                .iter()
+                .map(|child| child.render(bargs.eww_state, bargs.window_name, bargs.widget_definitions));
             // we know that we have exactly three children here, so we can unwrap here.
             let (first, center, end) = children.next_tuple().unwrap();
             let (first, center, end) = (first?, center?, end?);
--- a/crates/eww/src/widgets/widget_node.rs
+++ b/crates/eww/src/widgets/widget_node.rs
@@ -113,7 +113,7 @@ pub fn generate_generic_widget_node(
 ) -> AstResult<Box<dyn WidgetNode>> {
     if let Some(def) = defs.get(&w.name) {
         if !w.children.is_empty() {
-            return Err(AstError::TooManyNodes(w.children_span(), 0).note("User-defined widgets cannot be given children."))
+            return Err(AstError::TooManyNodes(w.children_span(), 0).note("User-defined widgets cannot be given children."));
         }
 
         let mut new_local_env = w


RianFuro:
Seems like my system is not representative... I've tried with qtile, xmonad, dwm and awesome; none of them render docked widgets correctly, neither without, nor with this pr included. Undocked widgets work well though.

Using Xfce4 renders the docked widget correctly. I've also experienced the performance problems using upstream and not with this pr included.

crates/eww/src/app.rs
```diff
@@ -362,29 +363,35 @@ fn apply_window_position(
...
+    let gdk_window = window.window().context("Failed to get gdk window from gtk window")?;
+    window_geometry.size = Coords::from_pixels(window.size());
     let actual_window_rect = get_window_rectangle(window_geometry, monitor_geometry);
-    gdk_window.move_(actual_window_rect.x, actual_window_rect.y);
+
+    let size = gdk_window.origin();
+
+    if actual_window_rect.x != size.1 || actual_window_rect.y != size.2 {
+        gdk_window.move_(actual_window_rect.x, actual_window_rect.y);
+    }
```
elkowar, line 374: `origin` does not return the size, but the position of the window
comparing the position to the size would also not make sense - the naming here is thus _really_ weird.

elkowar:
additionally, does this not cause a short flicker when opening the window?

I'd love to just say "this is a bug in i3, as the EWMH spec clearly states that WMs are expected to react to window properties changing at any point in time, which it clearly doesn't" - but given how popular i3 is, I fear this might not really be reasonable. However, if noone has done it yet, filing an issue about the bad spec-compliance with i3 would most likely be the cleaner solution (assuming that it would get fixed)

crates/eww/src/app.rs
RianFuro: true, I missed renaming that, thanks for the heads up :sweat_smile:

RianFuro:
> additionally, does this not cause a short flicker when opening the window?

I assumed that as well, but after using this pr for well over a week now on my working machine I haven't noticed any flicker.

> but given how popular i3 is, I fear this might not really be reasonable. 

Also something I assumed, but my solution is actually based on this gist: https://gist.github.com/Airblader/d4fe0088b05f720fb4a3291cd9ec5e9a
From the maintainer of the i3-gaps fork... so this circumstance seems to be at least known? I must admit I haven't checked if there's an open issue regarding spec-compliance and if you're really reluctant on integrating this workaround this would be my next avenue, but given this solution seems to be accepted I don't have high hopes honestly :sweat_smile:

elkowar:
I've filed an issue at https://github.com/i3/i3/issues/4510, could you take over the communication there? (I don't use i3, and I don't really want to have to set it up just for this ^^) @RianFuro

RianFuro:
Will do :)

Airblader:
FWIW, I wouldn't recommend merging such a change unless eww _intends_ to be able to change the window type at runtime (would be the first client I ever heard of doing that). The correct behavior is to set as many atoms as you can prior to mapping the window, especially those that aren't intended to change. Long discussions can be had over SHOULD vs MUST, but RFC 2119 defines SHOULD as

> […] that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course.

I would argue that in the world of X11, you'll have a really tough time as a client if you go against this. In any case, this is one of those implications to be understood if you want to keep the behavior. :-)

Airblader:
> Also something I assumed, but my solution is actually based on this gist:
> https://gist.github.com/Airblader/d4fe0088b05f720fb4a3291cd9ec5e9a

That script was a hack four years ago as well, and I only wrote it because someone made an intentional decision to go down this route. It is not something I'd ever _recommend_ anyone doing since there definitely are implications when a window becomes a dock client "against its will".

RianFuro:
> FWIW, I wouldn't recommend merging such a change

Yeah, I definitely agree with that now, thanks for clearing that up :)

crates/eww/src/app.rs
```diff
@@ -358,6 +359,9 @@ fn initialize_window(
         }
         display_backend::set_xprops(&window, monitor_geometry, &window_def)?;
     }
+
+    window.show_all();
+
```
RianFuro, line 364: So... it's actually pretty simple in hindsight :sweat_smile: 

Thanks again @Airblader for helping out, I learned something today :)

RianFuro:
@elkowar I think we're good now on that topic. I'm not sure if I still need the dependency update at this point; would you want to keep that? Otherwise I'll try to pull that out again.

elkowar:
I'd love to keep it, although it should probably be put into a separate PR

RianFuro:
Alright, this is just the fixes and a `cargo fmt` now. I'll put up the pr with the dependency updates later

elkowar:
This looks good now! thanks!

PR #259, opened 2021-09-03
Indicate label truncation

### LINKED ISSUE #257: [FEATURE] label should be able to indicate when it is truncated

Truncated labels (using `:limit-width`) are currently cut. I want to be able to see when the label has been truncated.

Add a `:limited-indicator` (or similar) attribute, which allows the user to pass a custom string which gets appended if the string was limited.

If this makes sense then I can look into implementing this.

Cheers!

ThreeFx:
Closes #257

I am not sure if the documentation is automatically updated when adding new parameters, can someone help me there? Thanks!

## Description

When `:show-truncated` and `:limit-width` are set, indicate label truncation by appending `...` to the displayed text.

## Usage

```lisp
(label :limit-width 20 :show-truncation true "012345678901234567890123456789")
```

### Showcase

image

## Checklist


elkowar:
Thanks!

PR #261, opened 2021-09-05
fix: correct gt and lt parsing

RianFuro:
The parser swapped the meaning of > and <, this pr corrects that.


elkowar (approved):
I am stupid, haha! Thanks!

PR #264, opened 2021-09-06
Makes code more rusty/idiomatic

legendofmiracles:
Most notable one in my opinion is the `expect` ⇒ `unwrap_or_else`, the reason being that expect will always call the function - only using the result when needed (so impure)

There are stilll some clippy lints, but at the current time most of them are false positives or just wrong (they produce compiler errors when applied)

### DIFF
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -153,7 +153,7 @@ impl App {
                     sender.respond_with_result(result)?;
                 }
                 DaemonCommand::CloseWindows { windows, sender } => {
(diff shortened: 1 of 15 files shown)


VuiMuich:
> There are stilll some clippy lints, but at the current time most of them are false positives or just wrong (they produce compiler errors when applied)

Might be a result of building with nightly rust, right?
So maybe reporting this to the floppy repo would be nice.

legendofmiracles:
Thank you, I can reproduce this with the latest nightly - but the version in the nix flake still works

elkowar:
Thanks!

PR #268, opened 2021-09-09
Fix variables to be scoped from literal created/called widgets

### LINKED ISSUE #267: [BUG] Variables aren't scoped from widget created or called by literal.
## Checklist before submitting an issue
- [x] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [x] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [x] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

## Describe the bug | Reproducing the issue

Sample self-contained repro `eww.yuck`:

```yuck
(defvar active_workspace 0)
(defvar workspaces '(box (workspace :index 0 :name "1"))')

(defwidget workspace [index name]
  (button :class "workspace ${active_workspace == index ? 'active' : 'inactive'}" :onclick {"wmctrl -s " + index}
    name))

(defwidget workspaces_lit []
  (literal :content workspaces))
(defwidget workspaces_manual []
  (box (workspace :index 0 :name "1")))

(defwindow topbar
  :monitor 0 :stacking "fg" :wm-ignore false :windowtype "dock"
  :reserve (struts :side "top" :distance "30px")
  :geometry (geometry :x "0" :y "0" :width "100%" :height "30px" :anchor "top center")
  (box (workspaces_lit) (workspaces_manual)))
```

Unknown variable error happens with `workspace_lit` widget (but `workspaces_manual` works just fine):
```
error: Unknown variable active_workspace
  ┌─ /home/animesh/TrashThings/eww-testing/literal-bug/eww/eww.yuck:5:31
  │
5 │   (button :class "workspace ${active_workspace == index ? 'active' : 'inactive'}" :onclick {"wmctrl -s " + index}
  │                               ────────────────
  │
  → Hint: If you meant to use the literal value "active_workspace", surround the value in quotes
```

## Expected behaviour
Variable should be able to be scoped from literal called/created widgets.

Animeshz:
## Description

Instead of passing empty `EwwState` (which I assume is added by mistake in staging incomplete code), passes the actual state by creating second mutable reference to the EwwState passed (because the `resolve_block` macro used the `bargs` mutably once and the actual mutation only happens after the block runs so it was safe operation to do it).

Closes #267

## Checklist

- [x] I used `cargo fmt` to automatically format all code before committing.

### DIFF
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -268,7 +268,6 @@ impl App {
         // refresh script-var poll stuff
         self.script_var_handler.stop_all();
 
-
         log::trace!("loading config: {:#?}", config);
 
         self.eww_config = config;
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -595,14 +595,12 @@ fn build_gtk_literal(bargs: &mut BuilderArgs) -> Result<gtk::Box> {
                 };
 
                 let widget_node = widget_node_result.context_label(literal_use_span, "Error in the literal used here")?;
-                let child_widget = unsafe {
-                    widget_node.render(std::mem::transmute(eww_state), &window_name, &widget_definitions)
+                let child_widget = widget_node.render(unsafe { std::mem::transmute(eww_state) }, &window_name, &widget_definitions)
                         .map_err(|e| AstError::ErrorContext {
                             label_span: literal_use_span,
                             context: "Error in the literal used here".to_string(),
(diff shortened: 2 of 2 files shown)


elkowar:
this would now set up new state-change -> widget change links into the main state constantly, leaving old ones intact, meaning this would pose a significant memory leak. additionally, I'm convinced that the use of unsafe should not be necessary here.

the main issue with literals referencing the state is that currently, any state linking is resolved prior to the initial render, meaning that dynamically generated links from state to widget aren't really possible. a rework of EwwState to make that said thing possible is planned and probably gonna be the next bigger step of development I'll work on.

Issue #241, opened 2021-08-22
[FEATURE] poll variables are blocking 


druskus20:
## Description of the requested feature
When opening a window, all poll variables are calculated once before opening. This can be slow specially when running commands or scripts that connect to the network. (i.e. package updates). In order to avoid this, variables could be evaluated asyncronously at start although they would probably require an initial value (similarly to listen variables).

## Proposed configuration syntax
```
(defpoll :initial "initial-value" :interval "1m" "myscript.sh")

druskus20:
Might try to implement this myself

druskus20:
I guess I have to now ^^

PR #272, opened 2021-09-11
Implement optional initial value for pollvars

druskus20:
## Description

This PR adds the functionality described in #241. It implements an optional "initial-value" field for pollvars. 

## Usage

When opening a window, all pollvars are calculated. By using initial-value when defining a pollvar, eww wont have to wait until the result of said pollvar is computed, therefore, making it not block the startup of the window.

An example of this can be found bellow.
```
; This pollvar blocks eww for 10s on startup.
(defpoll date_poll_blocking :interval "1m" "sleep 10 && date +%d/%m/%Y")

; This pollvar displays "initial_date" for 10s, thus, not blocking eww
(defpoll date_poll :initial "initial_date" :interval "1m" "sleep 10 && date +%d/%m/%Y")

; This pollvar gets calculated quick enough so it doesnt have time to display "initial_time"
(defpoll time_poll :initial "initial_time" :interval "1s" "date +%H:%M")
```

This PR shouldnt be a breaking change, as initial-value is an optional field, and, in its absence, the program should behave exacly as it did before.

## Checklist

### DIFF
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -268,7 +268,6 @@ impl App {
         // refresh script-var poll stuff
         self.script_var_handler.stop_all();
 
-
         log::trace!("loading config: {:#?}", config);
 
         self.eww_config = config;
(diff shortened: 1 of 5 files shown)


elkowar:
epic!

PR #273, opened 2021-09-11
Implement trigger for defpoll variable

### LINKED ISSUE #271: [FEATURE] Trigger for defpoll variable?

Sometimes we want some objects to be hidden over a revealer and only want to update the value of some defpoll variable when 
the revealer opens, in this case a trigger property (to false) may disable updates and update when its triggered and subsequently with the poll time until its trigger is back off.

This makes operation much more power efficient, specially when the updating interval is very small.

```yuck
(defpoll var_name :interval "0.3s" :trigger some_var_or_eww_expr
  `polling_cmd`)
```

Animeshz:
## Description

Adds trigger property to `(devpoll` variable, so that polling may stop when required and start at any instant with subsequent interval given.

Close #271

## Usage

```yuck
(defpoll :trigger <s-expr>
         :interval "0.2s"
         :initial "Initial value"
         `cmd`)
```

### Showcase

Sample widget controlled to be polled by another variable.

https://user-images.githubusercontent.com/26714676/132949323-24d69090-beb4-46ca-8505-e39a9865a101.mp4

## Additional Notes
\-

## Checklist

### DIFF
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -205,10 +205,10 @@ impl App {
     fn update_state(&mut self, fieldname: VarName, value: DynVal) {
         self.eww_state.update_variable(fieldname.clone(), value);
 
-        if let Some(linked_poll_vars) = self.eww_config.poll_var_trigger.get(&fieldname) {
+        if let Ok(linked_poll_vars) = self.eww_config.get_poll_var_link(&fieldname) {
             linked_poll_vars.iter().filter_map(|name| self.eww_config.get_script_var(name).ok()).for_each(|var| {
                 if let ScriptVarDefinition::Poll(poll_var) = var {
-                    match poll_var.trigger_expr.eval(self.eww_state.get_variables()).map(|v| v.as_bool()) {
+                    match poll_var.run_while_expr.eval(self.eww_state.get_variables()).map(|v| v.as_bool()) {
                         Ok(Ok(true)) => self.script_var_handler.add(var.clone()),
                         Ok(Ok(false)) => self.script_var_handler.stop_for_variable(poll_var.name.clone()),
                         Ok(Err(err)) => error_handling_ctx::print_error(anyhow!(err)),
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -12,7 +12,10 @@ use simplexpr::dynval::DynVal;
 use std::collections::{HashMap, HashSet};
 use tokio::sync::mpsc::UnboundedSender;
 use yuck::{
-    config::window_geometry::{AnchorPoint, WindowGeometry},
+    config::{
+        script_var_definition::ScriptVarDefinition,
+        window_geometry::{AnchorPoint, WindowGeometry},
+    },
     value::Coords,
 };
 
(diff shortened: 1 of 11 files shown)


crates/simplexpr/src/ast.rs
```diff
@@ -86,6 +86,39 @@ impl SimplExpr {
     pub fn synth_literal<T: Into<DynVal>>(s: T) -> Self {
         Self::Literal(s.into())
     }
+
+    pub fn dfs_collect_var_refs_into(&self, dest: &mut Vec<VarName>) {
```
elkowar, line 90: what does the `dfs` stand for here 🤔 just `collect_var_refs_into` should be enough

crates/eww/src/app.rs
```diff
@@ -200,7 +203,20 @@ impl App {
...
+                if let ScriptVarDefinition::Poll(poll_var) = var {
+                    match poll_var.trigger_expr.eval(self.eww_state.get_variables()).map(|v| v.as_bool()) {
+                        Ok(Ok(true)) => self.script_var_handler.add(var.clone()),
+                        Ok(Ok(false)) => self.script_var_handler.stop_for_variable(poll_var.name.clone()),
+                        Ok(Err(err)) => error_handling_ctx::print_error(anyhow!(err)),
+                        Err(err) => error_handling_ctx::print_error(anyhow!(err)),
+                    };
+                }
+            });
+        }
```
elkowar, line 219: this is somewhat ugly - It should be easily possible to flatten the structure here by using and_then to not have the nested results. Also, there is not all that much reason to go with for_each over a simple for-loop here

crates/eww/src/config/eww_config.rs
```diff
@@ -24,11 +24,18 @@ pub struct EwwConfig {
...
 
 impl Default for EwwConfig {
     fn default() -> Self {
-        Self { widgets: HashMap::new(), windows: HashMap::new(), initial_variables: HashMap::new(), script_vars: HashMap::new() }
+        Self {
+            widgets: HashMap::new(),
+            windows: HashMap::new(),
+            initial_variables: HashMap::new(),
+            script_vars: HashMap::new(),
+            poll_var_trigger: HashMap::new(),
```
elkowar, line 37: the name `poll_var_trigger` is reeaaaally confusing, see also my general comment on the name "trigger"

crates/yuck/src/config/script_var_definition.rs
```diff
@@ -71,10 +73,17 @@ impl FromAstElementContent for PollScriptVar {
             let initial_value = Some(attrs.primitive_optional("initial")?.unwrap_or_else(|| DynVal::from_string(String::new())));
             let interval = attrs.primitive_required::<DynVal, _>("interval")?.as_duration()?;
             let (script_span, script) = iter.expect_literal()?;
+
+            let trigger_expr =
+                attrs.ast_optional::<SimplExpr>("trigger")?.unwrap_or_else(|| SimplExpr::Literal(DynVal::from(true)));
```
elkowar, line 78: I really dislike the name `trigger`, as that makes it sound like it starts, but never stops the poll-var. The attribute would be better described by a name such as `only-when` or `while`.

elkowar (changes requested):

crates/simplexpr/src/ast.rs
Animeshz: `dfs` stands for `Depth-first search` just a name to define what type of traversal going to happen while collecting the refs, though is not necessary to know cuz its just implementation detail, I'll just rename it 👍.

crates/eww/src/app.rs
Animeshz: I don't think it'll remove any ugliness, `match` expr does this with 2 lines and 2 line for two different errors. Using `map_err` will only map the inner error, and outer error still need to be mapped using `anyhow!`. Also since line is long, if-else in the `and_then` will take 4-5 lines anyway 🤔🤔.

Animeshz: `and_then` won't accept this as `ConversionError` is different than `EvalError` and incompatible, it'll involve anyhow inside and once outside, increasing complexity of reading the code itself. I think its pretty explicit and readable right now 🤔🤔

crates/yuck/src/config/script_var_definition.rs
Animeshz: How about 3 letter `:run [true|false]`?

Animeshz: or `:run-while` to be more explicit?

crates/eww/src/app.rs
elkowar: true, in that case this is good enough for now 👍🏻

crates/yuck/src/config/script_var_definition.rs
elkowar: I like `:run-while`! the fields in the code should then also be renamed respectively.

crates/eww/src/config/eww_config.rs
elkowar: the fact that "trigger" is singular here also really doesn't make sense. `poll_var_run_while_dependencies` or something would probably be a better name,... somewhat unweildly, but ehhh, it's a somewhat confusing edge-case so better be descript imo

Animeshz:
👍

elkowar:
This looks good now! thanks a lot!

PR #274, opened 2021-09-11
gtk dependency update to 0.14

RianFuro:
Update dependencies of gtk, gdkx11, and cairo-sys to 0.14

Structurally, nothing has really changed, I couldn't observe any breaking changes due to the update. Changes are exclusively naming adaptions, as in one of their updates, they changed their naming conventions to more closely align with rust's general naming conventions.


elkowar:
Looks good, thanks!

PR #276, opened 2021-09-13
Exclude duplicates of widget property in docs

Animeshz:
## Description

When same property is used multiple times in the [widget_definition.rs](https://github.com/elkowar/eww/blob/master/crates/eww/src/widgets/widget_definitions.rs) and annotated with `@prop` for giving description, the docs generated writes it multiple times.

Previously the `timeout` was shown twice, now by addition of `onhoverlost` using the same property (timeout) again, it is shown thrice 😂. This simply exclude duplicate property by same name in the docs.


elkowar:
Nice, thank you!

PR #297, opened 2021-10-05
Wayland support in flake.nix

rprospero:
## Description

Provide a short description of the changes your PR introduces.
This includes the actual feature you are adding,
as well as any other relevant additions that were necessary to implement your feature.

This pull request adds a second package to `flake.nix` with Wayland support.

## Usage

When adding a widget or anything else that affects the configuration,
please provide a minimal example configuration snippet showcasing how to use it and

It will be possible to get a wayland enabled `eww` via `nix build github:elkowar/eww#eww-wayland`, just as it is possible to get an X11 enabled version via `nix build github:elkowar/eww`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,8 +31,10 @@
 
         eww-wayland = naersk-lib.buildPackage {
           pname = "eww";
+          buildInputs = with pkgs; [ gtk-layer-shell ];
           nativeBuildInputs = with pkgs; [ pkg-config gtk3 ];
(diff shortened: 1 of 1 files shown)


legendofmiracles:
Thank you for your contribution, I did something similar a while back but never upstreamed it, and I honestly find it a bit cleaner: https://github.com/legendofmiracles/eww/blob/e2c4ac31271b0daaef70ccc2e0670216824ee77d/flake.nix#L24
Maybe you can also use a builder function to reduce duplicated code?

Issue #233, opened 2021-08-17
[FEATURE] A circle progress bar


SkyLissh:
## Description of the requested feature
I try make it by myself, just with css and xml, but no luck. While looking if someone tried the same, I found a repo of the implementation with GTK in vala.

## Proposed configuration syntax
May use
```xml
 <progress circle="true" />
```

# Additional context
![image](https://user-images.githubusercontent.com/72633267/129667503-7355095c-f504-4077-9838-95d67a68e27e.png)
Repo of [phastmike](https://github.com/phastmike/vala-circular-progress-bar)

undefinedDarkness:
Oooo, that is a nice example

druskus20:
working on this woo

legendofmiracles:
Will be closed by #328 (read: elkowar please link lol)

elkowar:
Was indeed implemented by #328!

PR #328, opened 2021-10-31
Circular progress widget

druskus20:
## Description

This PR implements circular progress widgets as mentioned in #233. 

## Usage

```lisp 
(box :class "bar"
     :space-evenly true
  (circular-progress :class "thingy"
                   :value {round(EWW_CPU.avg, 0)}
                   :thickness 5
                   :clockwise true
                   :start-at 30
         (label :text "${round(EWW_CPU.avg, 0)}"))))
```

### Showcase
image

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -63,7 +63,7 @@ checksum = "a83b21d2aa75e464db56225e1bda2dd5993311ba1095acaa8fa03d1ae67026ba"
 dependencies = [
  "atk-sys",
  "bitflags",
- "glib 0.14.4",
+ "glib 0.14.8",
  "libc",
 ]
 
@@ -152,7 +152,7 @@ checksum = "f859ade407c19810ae920b4fafab92189ed312adad490d08fb16b5f49f1e2207"
 dependencies = [
  "bitflags",
  "cairo-sys-rs",
- "glib 0.14.4",
+ "glib 0.14.8",
  "libc",
  "thiserror",
 ]
@@ -428,7 +428,7 @@ dependencies = [
  "gdk-pixbuf 0.9.0",
  "gdkx11",
  "gio 0.9.1",
- "glib 0.10.3",
+ "glib 0.14.8",
  "grass",
  "gtk",
  "gtk-layer-shell",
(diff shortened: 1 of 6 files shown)


druskus20:
This should be done now

druskus20:
So basically, the way to center something inside the widget currently is: 
```
halign: "center"
valign: "center"
height: 20
width: 20
```
Because there's currently an issue with setting both height and width at the same time (being fixed in state_rework

druskus20:
Woops

druskus20:
Currently there is an issue with adding multiple children to a `circular-progress` widget. Because it is a bin, it only allows one direct children. If the user tries to ad more, gtk will error out as it is a configuration error: 
```
(eww:143555): Gtk-WARNING **: 13:42:03.247: Attempting to add a widget with type GtkLabel to a CircProg, but as a GtkBin subclass a CircProg can only contain one widget at a time; it already contains a widget of type GtkLabel
```

However, it doesnt end there, and gtk continues to throw draw() errors every redraw:
```
(eww:140854): Gtk-CRITICAL **: 13:38:55.443: gtk_container_propagate_draw: assertion '_gtk_widget_get_parent (child) == GTK_WIDGET (container)' failed
```

This in my opinion just makes it harder for the user to figure out what the problem is. The only way to prevent those messages that I can come up with is adding an additional check at the end of `CircProgPriv`'s `draw` function.
```rust
// Old
if let Some(child) = &*self.content.borrow() {
    widget.propagate_draw(child, &cr); 
}
// New
if let Some(child) = &*self.content.borrow() {
    if let Some(_parent) = child.parent() { widget.propagate_draw(child, &cr); }
}
```

druskus20:
> So basically, the way to center something inside the widget currently would be:
> ....

This has been fixed by using cairo clip, and it will also allow to set a background color in the children widget.



> Currently there is an issue when the user tries to have multiple children inside a `circular-progress` widget. Because it is a subclass of Gtk::Bin, it will only allow for one direct children. Therefore, gtk with error out. This is intended.

This has also been fixed by checking that the container doesnt already have a children at the moment of inserting a new one.

druskus20:
At this point it works. However, there's still the issue that gtk::Bin doesnt support margin / padding. This is also relevant for #338

druskus20:
Margin right now should behave like it does with a normal `gtk::Box`

PR #334, opened 2021-11-04
Update README.md with another example setup

iSparsh:
## Description
Added link to my repo as well as showcase image in the README for examples.

## Usage
Helps people understand structured yuck configs, (rudimentary) rofi-like configuration example as well as power-menu popup

## Showcase
Added showcase image (all 3 windows)

## Additional Notes
I had a few questions/suggestions
Should we create a README in the examples directory which has all the example links and pictures (like this one) over there, instead of populating the main README?
Let me know about any changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ Additionally, a couple _amazing_ people have started to work on an
 * [Setup by adi1090x](https://github.com/adi1090x/widgets)
 ![Nordic](https://raw.githubusercontent.com/adi1090x/widgets/main/previews/nordic.png)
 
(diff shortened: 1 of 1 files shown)


elkowar:
Looks good! I think it'd be good if you renamed the eww_windows etc files to actually be `.yuck`, such that people cloning it actually have the correct editor support etc if they have that set up. 

Putting examples into a separate file might soon be a good idea, as more and more examples accumulate - but for now, this is pretty good

iSparsh:
@elkowar done! thanks so much. I'll be sure to make more widgets. PS: Excited for the state rework and new graphs 👀

PR #342, opened 2021-11-08
Remove superfluous period in an inline code block

Vermoot:
There's a period there that shouldn't be there.

Am I a contribewwtor or WHAT


elkowar:
Hell yea, thanks!

Issue #332, opened 2021-11-03
[FEATURE] Proper Magic Variable Documentation


iSparsh:
## Description of the requested feature
Not enough documentation for the eww magic variables. There is no way to know which values to use on the wiki. For example: `EWW_CPU` has a `.avg` value that can be accessed, but I only know that through example configurations. There is no easy way to know these values easily on the Wiki page.

legendofmiracles:
Can this be assigned to me?

legendofmiracles:
@iSparsh This is what I got so far, quick and ugly. 
![image](https://user-images.githubusercontent.com/30902201/140592753-38aeb28a-a49a-4586-9950-24fab2addf39.png)
Anything that you don't like?

iSparsh:
That's perfect @legendofmiracles ! Thanks so much.

legendofmiracles:
That was my plan originally - when i first added them, but elkowar didn't like it because they would have to be maintained. 
The one example that exists only exists because I wanted to showcase the special syntax.

iSparsh:
Alright then! I guess its resolved. Thanks!

legendofmiracles:
Uhh, I haven't opened the PR yet.
I'd rather have it open to remind to finish it

PR #343, opened 2021-11-08
add properties to magic vars

legendofmiracles:
Closes https://github.com/elkowar/eww/issues/332


elkowar:
might be more fitting to call the heading "structure" or something, has having it be "properties" and then just a single json codeblock is kinda weird

elkowar:
looking good, thanks!

Issue #355, opened 2021-11-17
[FEATURE] Add watcher for variable changes


owenrumney:
## Description of the requested feature
I want to be able to watch a variable for a deflisten and be able to update another variable. The example I'm thinking is when playerctl catches a track change, I want to do an 
`eww update revealSpotify=${revealSpotify ? false : true}`

I think this has the potential to be useful in other use cases

## Proposed configuration syntax

I'm thinking something along the lines of 
```
# capture the changes of tracks 
(deflisten music :initial "" "playerctl --follow metadata --format '{{ artist }} - {{ title }}' || true")

# monitor the listen to perform the command on state change
(defwatcher name :watch music  :initial "false" "eww update reveal=${reveal ? false : true}")
```

# Additional context

I'm using this to learn a bit of Rust and I'm happy to contribute - however, if you don't think its useful or have something in the pipeline that is going to fulfil this then I should probably look at something else.

owenrumney:
I'm wondering if I've over-thought this - maybe the answer here is much simpler and add an `onChange` to `defListen`
```
(deflisten music
  :initial ""
  :onChange "eww update revealSpotify=true && sleep 5 && eww update revealSpotify=false"
"playerctl --follow metadata --format '{{ artist }} - {{ title }}' || true")
````

owenrumney:
Withdrawing as per contents of #356

PR #356, opened 2021-11-17
Add onchange to deflisten

owenrumney:
## Description

Adding onchange to deflisten as per #355 so that I can pop out the spotify expander.

I think this has other uses for things like a deflisten that watches for github stars or PRs and triggers something on the machine.

## Usage

Add an `onchange` attribute to the `deflisten` with the command you want to run.

```
(deflisten music
  :initial ""
  :onchange "eww update revealSpotify=true && sleep 5 && eww update revealSpotify=false"
"playerctl --follow metadata --format '{{ artist }} - {{ title }}' || true")
```

### Showcase

listenonchange

## Additional Notes

This is Day0 of writing Rust, if it is totally off the mark or not Rustic then I can change as required.

I ran `cargo fmt` and it changed loads of files :eyes:

## Checklist


elkowar:
I don't quite see the utility here, when you could also simply do
```
playerctl --follow metadata --format '{{ artist }} - {{ title }}' | while read -r line; do
  echo $line
  { eww update revealSpotify=true; sleep 5; eww update revealSpotify=false } &
done
```
without requiring any specific support form eww

owenrumney:
Yeah, that's valid I guess! I also use it with a small utility listening for dbus notifications and it slides out an expander with the message. From a usage perspective, I think it's more clear that when the variable changes something explicit will happen, rather than all wrapped in the listening command.

TL;DR; I take your point, happy for you to reject if you don't see a place for it in upstream.

elkowar:
Yea, I'm afraid I don't think this is really necessary, and would probably add more confusion than anything else. In the future, there will be some changes that allow you to change ewws state more directly from within eww, in which case this would maybe be a useful addition. Until that is in place however I don't think this provides enough utiIty. Glad you understand, and still very much thanks for your contribution!

PR #379, opened 2022-01-03
Added gtk-layer-shell as a dependency in the docs

### LINKED ISSUE #370: Add `gtk-layer-shell` to dependency list in docs
Currently, `eww` depends upon [gtk-layer-shell](https://github.com/wmww/gtk-layer-shell) when compiled with `--features=wayland`.  It might be helpful to list this in the [documentation](https://elkowar.github.io/eww/eww.html) under "Packages".

Robert7301201:
gtk-layer-shell is a required dependency but is not listed in the docs. This pull request adds gtk-layer-shell to the list of dependencies in the docs.

Closes #370

### DIFF
--- a/docs/src/eww.md
+++ b/docs/src/eww.md
@@ -27,6 +27,7 @@ The following list of package names should work for arch linux:
 <summary>Packages</summary>
 
 - gtk3 (libgdk-3, libgtk-3)
+- gtk-layer-shell
 - pango (libpango)
 - gdk-pixbuf2 (libgdk_pixbuf-2)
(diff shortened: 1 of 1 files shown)


elkowar:
Ohhh true, I forgot the Wayland specific deps when setting these up. Would probably be good to note that it's only required on Wayland, I. E.
`gtk-layer-shell (only on Wayland)`

Robert7301201:
Updated docs to parenthetically note that `gtk-layer-shell` is only required on Wayland.

elkowar:
Thanks!

PR #389, opened 2022-02-02
Add new vertical bar examples.

rxyhn:
I've added a aesthetic vertical bar example for eww documentation.


elkowar:
Thanks, this looks great!

PR #396, opened 2022-02-06
fix: limit ipc socket filename length

viandoxdev:
## Description

This PR changes how the ipc socket file is named to make sure that it doesn't exceed the 108 bytes limit for socket filenames on linux, which could be hit with a long configuration path.

This is done by changing how `daemon_id` (variable dictating the eww log and ipc socket filenames) is made: by hashing the config directory path instead of base64 encoding it.

### Additional Notes

This also adds a warning when the path still ends up longer than 100 bytes, in the (**very**) rare case that someone changed their `XDG_RUNTIME_DIR` to something pretty long.

## Checklist

> This is probably not a notable change.

### Tested

 - [x] compiling
 - [x] running the daemon
 - [x] sending commands (opening and closing windows)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -412,7 +412,6 @@ name = "eww"
 version = "0.2.0"
 dependencies = [
  "anyhow",
- "base64",
  "bincode",
  "cairo-rs",
  "cairo-sys-rs",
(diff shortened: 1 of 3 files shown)


elkowar:
This looks good! thanks!
Given that this seems to be the only use of the base64 crate, can we remove that dependency as well, now that it's unused?

elkowar:
Looks good! Thanks!

Issue #394, opened 2022-02-05
[FEATURE] Way to access JSON arrays length


viandoxdev:
### Description of the requested feature

Not sure if I'm being dumb, but AFAIK, there is currently no way to access a JSON array's length in yuck, even though yuck does support the arrays themselves, so I'd like a way to do exactly that.

### Proposed configuration syntax

How it's done doesn't really matter but here's a few ways:

the javascript way:
```js
array.length
```
or in a function
```js
len(array)
```
 if done this way (function) it could also return strings' length, and be useful to know if a string would fit in a certain space or not (and replace it with another if it doesn't).

### Additional context

This would be useful when trying to display variable length JSON data. In my case I'm trying to display contributions obtained from the Github API.

### Workarround

For now since I'm already using jq to simplify the API's responses, I can do:
```bash
... | jq '{length: length, data: .}'
```
which gives
```json
{
    length: ...,
    data: [...]
}
```
and use it in yuck like this
```yuck
{ arr.data[arr.length - 1] }
```

Patch that implements this feature:
```diff
diff --git a/crates/simplexpr/src/eval.rs b/crates/simplexpr/src/eval.rs
index a3f1789..c9a3def 100644
--- a/crates/simplexpr/src/eval.rs
+++ b/crates/simplexpr/src/eval.rs
@@ -273,6 +273,13 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
             }
             _ => Err(EvalError::WrongArgCount(name.to_string())),
         },
+        "length" => match args.as_slice() {
+            [string] => match string.as_json_value().as_ref().map(|x| x.as_array()) {
+                Ok(Some(vec)) => Ok(DynVal::from(vec.len() as i32)),
+                _ => Ok(DynVal::from(string.as_string()?.len() as i32)),
+            },
+            _ => Err(EvalError::WrongArgCount(name.to_string())),
+        },
         _ => Err(EvalError::UnknownFunction(name.to_string())),
     }
 }

```

PR #407, opened 2022-02-23
Add several functions for simple expressions

MartinJM:
## Description

This PR adds a `search`, `captures`, and `with` function to the simple expressions.

- The `search` function will return all occurances of the regex. Return format is json, where the occurances are numbered.
- The `captures` function will return all captures of all the occurances of the regex. Return format is also json, but contains all capture groups separately.
- The `with` function allows for variables within simple expressions, which are only valid within the expression in the third argument.

The `search` and `capture` functions are so that there is less need for `grep` in `defpoll`s. This saves the creation of new processes.

The `with` function is mostly there because both the `search` and `captures` return multiple values (in a json object). To prevent having to do either multiple times to get multiple results, the `with` can be used.

## Usage

Simple examples:
```
search('abba', 'b')
{"0": "b", "1": "b"}

captures('abba', 'a(b)|ba')
{"0": {"0": "ab", "1": "b"}, "1": {"0": "ba"}}

with('a', '{"0": "hello ", "1": "world"}', 'a[0] + a[1]')
hello world
```

A real-world example in Eww (for volume):
```
(defwidget volume []
	(label
		:text "${ with('a', captures(master_volume, 'Left: .* \\[([[:digit:]]*%)\\] \\[(.*)\\]'), '(a[0][2] == \"off\"? \"Muted (\" : \"Volume: \") + a[0][1] + (a[0][2] == \"off\" ? \")\" : \"\")') }"
	)
)

(defpoll master_volume
	:interval "1s"
	:initial ""
	:run-while "true"
	`amixer sget Master`
)
```
Without the `with`, the whole `capture` would need to be in here 3 times.

## Additional Notes

If the `search` and `captures` functions are good enough for merging, but the `with` function is not, I can also make a separate PR for just them.

## Checklist

### DIFF
--- a/crates/simplexpr/src/eval.rs
+++ b/crates/simplexpr/src/eval.rs
@@ -285,17 +285,17 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
                 let string = string.as_string()?;
                 let pattern = regex::Regex::new(&pattern.as_string()?)?;
                 let mut retval = String::new();
-                retval.push_str("{");
+                retval.push_str("[");
                 let mut find_present = false;
-                pattern.find_iter(&string).enumerate().for_each(|(i, x)| {
+                pattern.find_iter(&string).for_each(|x| {
                     find_present = true;
-                    retval.push_str(&format!("\"{}\": \"{}\", ", i, x.as_str()));
+                    retval.push_str(&format!("\"{}\", ", x.as_str()));
                 });
                 if find_present {
                     retval.pop();
                     retval.pop();
                 }
-                retval.push_str("}");
+                retval.push_str("]");
                 Ok(DynVal::from(retval))
             }
             _ => Err(EvalError::WrongArgCount(name.to_string())),
@@ -305,17 +305,17 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
                 let string = string.as_string()?;
                 let pattern = regex::Regex::new(&pattern.as_string()?)?;
                 let mut retval = String::new();
-                retval.push_str("{");
+                retval.push_str("[");
                 let mut capture_present = false;
                 let mut inner_capture_present = false;
-                pattern.captures_iter(&string).enumerate().for_each(|(i, x)| {
+                pattern.captures_iter(&string).for_each(|x| {
                     capture_present = true;
                     inner_capture_present = false;
-                    retval.push_str(&format!("\"{}\": {{", i));
-                    x.iter().enumerate().for_each(|(j, y)| match y {
+                    retval.push_str("[");
+                    x.iter().for_each(|y| match y {
                         Some(m) => {
                             inner_capture_present = true;
-                            retval.push_str(&format!("\"{}\": \"{}\", ", j, m.as_str()));
+                            retval.push_str(&format!("\"{}\", ", m.as_str()));
                         }
                         None => {}
                     });
@@ -323,13 +323,13 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
                         retval.pop();
                         retval.pop();
                     }
-                    retval.push_str("}, ");
+                    retval.push_str("], ");
                 });
                 if capture_present {
                     retval.pop();
                     retval.pop();
                 }
-                retval.push_str("}");
+                retval.push_str("]");
                 Ok(DynVal::from(retval))
             }
             _ => Err(EvalError::WrongArgCount(name.to_string())),
--- a/crates/simplexpr/src/eval.rs
+++ b/crates/simplexpr/src/eval.rs
@@ -306,16 +306,19 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
             }
             _ => Err(EvalError::WrongArgCount(name.to_string())),
         },
-        "length" => match args.as_slice() {
-            [string] => match string.as_json_value().as_ref().map(|x| x.as_array()) {
-                Ok(Some(vec)) => Ok(DynVal::from(vec.len() as i32)),
-                _ => match string.as_json_value().as_ref().map(|x| x.as_object()) {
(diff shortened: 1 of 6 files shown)


MartinJM:
I made a mistake in my code (empty search/captures don't work), so will be fixing that.

MartinJM:
Should be fixed now.

viandoxdev:
If you want you could also add the length function I made. There's a patch for it (on #394) but I didn't want to make a PR for just one function.

MartinJM:
Thank you for the replies!

> If you want you could also add the length function I made. There's a patch for it (on #394) but I didn't want to make a PR for just one function.

Oh cool, I'll add that as well when I have a little more time. I also implemented a hex decode function, which I'll probably add as well then.


> I definitely think the search and captures functions would be a great addition to the expression syntax. I'm not sure if I like the name `search`, then again I can't think of a better name currently ^^'

I agree that the name is not ideal, maybe `find` would be better?

> I think both of those functions should definitely use arrays over json objects as their return value, i.e. `captures("2020-12, 2021-01", "(\d\d\d\d)-(\d\d)")` should return `[["2020", "12"], ["2021", "01"]]`.

This does make more sense indeed. I will change that.

> While I think something like the `with` function is definitely a great addition, I'd argue it should probably rather be some new syntax, over having a function for that. a classic in many languages would be something along the lines of
> 
> ```
> let matches = search("foo bar baz", "f..") in matches[0]
> ```
> 
> That probably warrants a separate PR or feature by me, though ^^
> [...]
> I'll create a new feature request for the let-in syntax though, such that that can be tracked separately.

I'll remove the `with` function from this PR and add the information to the feature request, so that it can be discussed independently from the other functions.

> I currently don't really have much time for reviewing the code here (see the maintainance status section in the readme), so I can't guarantee that this will go super fast, I'm afraid.

No worries at all, and good luck with your thesis :+1:

MartinJM:
Sorry, I wanted to rename the branch, which should work, but apparently doesn't.

MartinJM:
@viandoxdev I've added the `length` function. I did change it so that it matches the `jq length` on objects as well.

MartinJM:
This now contains the following functions:
 - `search`
 - `capture`
 - `hex_decode`
 - `length`
 
 The output of the `search` and `captures` functions has changed to arrays.
 
 Examples:
 ```
 search('2020-12, 2021-01', '([[:digit:]]{4})-([[:digit:]]{2})')
 ["2020-12", "2021-01"]
 
 captures('2020-12, 2021-01', '([[:digit:]]{4})-([[:digit:]]{2})')
 [["2020-12", "2020", "12"], ["2021-01", "2021", "01"]]
 
 hex_decode('48656c6c6f')
 Hello
 
 length('abc')
 3
 
 length('[1, 2, 3]')
 3
 
 length('{"1": "1", "2": "2"}')
 2
 ```

elkowar:
Ok, finally getting back to this! Looks mostly good now!

I was just made aware of the fact that having `length` work on both strings and json is rather weird here, as that means that if you for some reason have a string that accidentally contains valid json somewhere where you expect a string, you'll get the wrong length.
For that reason, I'd say it'd be good to make this two functions, one called `str_length` and one called `json_length`, which would separate the two -- str_length treating everything as a text, even if it's valid json, and json_length erroring out when provided with invalid json


Also: What's the actual usecase of hex_decode? seems like an oddly specific, edgecase-y thing to add

crates/simplexpr/src/eval.rs
```diff
@@ -276,6 +280,80 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
...
+                        retval.pop();
+                    }
+                    retval.push_str("], ");
+                });
+                if capture_present {
+                    retval.pop();
+                    retval.pop();
+                }
+                retval.push_str("]");
+                Ok(DynVal::from(retval))
```
elkowar, line 333:
```suggestion
                use serde_json::Value;
                let string = string.as_string()?;
                let pattern = regex::Regex::new(&pattern.as_string()?)?;
                Ok(Value::Array(
                    pattern
                        .captures_iter(&string)
                        .map(|captures| {
                            Value::Array(captures.iter().flatten().map(|x| Value::String(x.as_str().to_string())).collect())
                        })
                        .collect(),
                )
                .try_into()?)

```

crates/simplexpr/src/eval.rs
```diff
@@ -276,6 +280,80 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
...
+                pattern.find_iter(&string).for_each(|x| {
+                    find_present = true;
+                    retval.push_str(&format!("\"{}\", ", x.as_str()));
+                });
+                if find_present {
+                    retval.pop();
+                    retval.pop();
+                }
+                retval.push_str("]");
+                Ok(DynVal::from(retval))
```
elkowar, line 299:
```suggestion
                use serde_json::Value;
                let string = string.as_string()?;
                let pattern = regex::Regex::new(&pattern.as_string()?)?;
                Ok(Value::Array(pattern.find_iter(&string).map(|x| Value::String(x.as_str().to_string())).collect())
                    .try_into()?)

```

elkowar (changes requested):
In addition to the aforementioned changes to length, and me questioning the necessity of `hex_decode`, these would be some nicer ways to implement the search and captures functions

MartinJM:
Hey, thanks for the review. I don't have much time to work on this for a while, so I'm not sure when I'll be able to make the changes. You should be able to write on this branch as well, feel free to make changes as you see fit.

I agree with your comments on the `length`, `search`, and `captures` functions.

The `hex_decode` functionality is indeed for an edge-case. There are some tools that output hex values when they encounter unicode. With a hex decode function this can easily be changed back into unicode.

elkowar:
Should there not be some other way to make these tools output stuff in the expected manner? Having a hex_decode function just to work around some specific weird tools seems weird, especially given that it should be easily possible to worst-case just pipe the output of said tool into some other hex-decoding shell utility and do it shell-wise

elkowar:
I pushed some fixes and rebased onto master, so this should be ready to merge now 👍🏻

PR #425, opened 2022-04-04
EWW_BATTERY is empty string when its {total_avg} property would be NaN

### LINKED ISSUE #373: [BUG] when Linux has no batteries, EWW_BATTERY is not valid JSON

### Description of the bug

I use the same eww.yuck on both a gaming PC (no batteries), and a laptop: https://gitlab.com/jokeyrhyme/dotfiles/-/blob/main/config/eww/eww.yuck

I've noticed that on my PC, with no batteries, the EWW_BATTERY value is not valid JSON, which makes working with it tricky

### Reproducing the issue

```
$ eww state
EWW_BATTERY: , "total_avg": NaN}
EWW_RAM: {"total_mem": 33588690, "free_mem": 19586744, "total_swap": 240031621, "free_swap": 240031621, "available_mem": 27843256, "used_mem": 5745432, "used_mem_perc": 17.105259}
```

You can see here that EWW_RAM is valid JSON, but EWW_BATTERY is truncated, and I'm also not sure if `NaN` is a valid JSON number

### Expected behaviour

I'm not quite sure what to expect, I suppose this is a question of design

- should EWW_BATTERY be completely absent when there is no battery ?
- should it be an empty string ?
- should it be JSON like `{ "total_avg": null }` ?

### Additional context

I believe this line might need some adjustment: https://github.com/elkowar/eww/blob/master/crates/eww/src/config/system_stats.rs#L153

I might have an attempt at this myself if I get some free time

andornaut:
Fixes #373

I followed [the suggestion in this comment](https://github.com/elkowar/eww/issues/373#issuecomment-997439825) from [the linked issue](https://github.com/elkowar/eww/issues/373).

> IMHO it should just completely be null then, when there is no battery, there isn't much use to having the json values in the first place.

## Description

`EWW_BATTERY` will be `{ total_avg: ... }` or `""` if `total_avg` would otherwise have been `NaN`, which can occur if the path `/sys/class/power_supply` is empty.

## Usage

```
(defwidget battery []
  (box :class "battery"
    {EWW_BATTERY == "" ? "" : EWW_BATTERY.total_avg}))
```

## Additional Notes

I removed [`json.pop()`](https://github.com/elkowar/eww/blob/ee07598a039e191239a143407bf934b3781ee675/crates/eww/src/config/system_stats.rs#L152), because I believe it always removed "," only to add it back in the next line. which I removed as well.

## Checklist

### DIFF
--- a/crates/eww/src/config/system_stats.rs
+++ b/crates/eww/src/config/system_stats.rs
@@ -149,9 +149,11 @@ pub fn get_battery_capacity() -> Result<String> {
             }
         }
     }
-    json.pop();
-    json.push_str(&format!(r#", "total_avg": {:.1}}}"#, (current / total) * 100_f64));
+    if total == 0_f64 {
+        return Ok(String::from(""));
+    }
 
+    json.push_str(&format!(r#" "total_avg": {:.1}}}"#, (current / total) * 100_f64));
     Ok(json)
 }
 


elkowar:
wouldn't it actually make more sense to just pretend you always have a full battery, and thus go `total_avg: 100`? that'd avoid having to explicitly test for it being invalid json

andornaut:
> wouldn't it actually make more sense to just pretend you always have a full battery, and thus go `total_avg: 100`? that'd avoid having to explicitly test for it being invalid json

No. I'd rather not show the battery in that case. (I use the same configs on multiple computers).

elkowar:
Fair point, I guess. I'll merge this!

PR #440, opened 2022-04-26
[WIP] AttrValue / Update

druskus20:
## Description

**The PR** 

## Checklist


elkowar:
Still TODO:
- [ ] implement placeholders for update (and figure out how they should look) (curlies are pain because they actually could be empty json objects,... maybe remove unnumbered `{}` now?)
- [ ] allow for multiple children
- [ ] Possibly implement a `const` or sth that keeps state references linked
- [ ] update documentation
- [ ] update widget attribute documentation !!!!
- [ ] handle defaults for actions in def_widget

elkowar:
Update:
To _properly_ implement actions as part of the data structure system here, we need to completely restructure the project.
Because actions would need to be represented as simplexpr/dynvals, the simplexpr crate would need to not only _know_ about the datastructure, but also know how to _parse_ the s-expression based update syntax.

The alternative would be to turn this into some sort of actual lambda syntax in simplexpr, and then having a side-effecty simplexpr.... but that's also horrible, and would look significantly worse.

I'm not sure how best to do this.


The other alternative is to, for now, not allow passing actions around through custom widgets, storing them in state, etc. That's a big limitation, but the only way I can currently see how to do this without _massive_ changes to the underlying architecture.

Issue #443, opened 2022-04-28
[BUG] On value updated old signal bindings are not released from widgets causing previous handlers execution on trigger


stremlenye:
### Checklist before submitting an issue

- [X] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [X] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [X] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

### Description of the bug

In the configuration like 
```lisp
(button 
      :onclick "echo ${value}"
      "first button")
    (button 
      :onclick "echo ${second-value}"
      "second button")
```

updating the `value` variable causes the signal bindings not to be released causing the previously
bound handlers to be executed along with the most recent ones.

### Reproducing the issue

_eww.yuck_
```lisp
(defvar value "I - 1")
(defvar second-value "II - 1")


(defwindow test
  :monitor 0
  :geometry (geometry 
    :x "34%"
    :y "-40px"
    :width "250px"
    :height "230px"
    :anchor "bottom center")
  (box
    :orientation "v"
    :spacing 10
    (button 
      :onclick "echo ${value}"
      "first button")
    (button 
      :onclick "echo ${second-value}"
      "second button")
))
```

_test.sh_
```sh
eww -c ~/.config/eww kill && \
eww -c ~/.config/eww daemon && \
eww -c ~/.config/eww open test && \ 
eww -c ~/.config/eww update value="I - 2" && \
eww -c ~/.config/eww update second-value="II - 2" && \
eww -c ~/.config/eww update second-value="II - 3" && \
echo "done"
```

Then click on both buttons.

The logs would show:

```sh
(eww:1904136): GLib-GObject-WARNING **: 19:29:40.611: ../../../gobject/gsignal.c:2736: instance '0x55e02a33a340' has no handler with id '37'

(eww:1904136): GLib-GObject-WARNING **: 19:29:40.636: ../../../gobject/gsignal.c:2736: instance '0x55e02a33a180' has no handler with id '45'

(eww:1904136): GLib-GObject-WARNING **: 19:29:40.648: ../../../gobject/gsignal.c:2736: instance '0x55e02a33a340' has no handler with id '80'

I - 1
I - 2
II - 1
II - 3
```

[...]

PR #445, opened 2022-04-28
Fix/signal disconnect

### LINKED ISSUE #437: [BUG] Broken Signal handler stuff, probably because of drag-and-drop

### Description of the bug

Eww frequently outputs errors along the lines of
```
(eww:4709): GLib-GObject-WARNING **: 21:39:00.247: ../glib/gobject/gsignal.c:2731: instance '0x55f5c107f6c0' has no handler with id '2569'

(eww:4709): GLib-GObject-WARNING **: 21:39:00.247: ../glib/gobject/gsignal.c:2731: instance '0x55f5c107f500' has no handler with id '2578'

(eww:4709): GLib-GObject-WARNING **: 21:39:00.247: ../glib/gobject/gsignal.c:2731: instance '0x55f5c107f340' has no handler with id '2587'

(eww:4709): GLib-GObject-WARNING **: 21:39:00.247: ../glib/gobject/gsignal.c:2731: instance '0x55f5c107fa40' has no handler with id '2596'
```
which occur only (or at least a lot more frequently) since we added drag and drop

### Reproducing the issue

Any config should work

### Expected behaviour

no errors

### Additional context

_No response_

stremlenye:
## Description

Fix for #443 
Replaces the static variable with the widget internal data field for SignalHandleId storage

## Additional Notes
Fixes #437 as well

## Checklist

### DIFF
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -13,6 +13,7 @@ use eww_shared_util::Spanned;
 use gdk::{ModifierType, NotifyType};
 use glib::signal::SignalHandlerId;
 use gtk::{self, glib, prelude::*, DestDefaults, TargetEntry, TargetList};
+use glib::translate::FromGlib;
 use itertools::Itertools;
 use once_cell::sync::Lazy;
 use std::hash::Hasher;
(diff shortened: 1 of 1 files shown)


stremlenye:
> You'd need to use unique id's, no?
I've just figured it exactly that :sweat_smile:  
Will try to fix that today.

elkowar:
I'm curious, with current master, does the issue this fixes actually still occur? I very recently also did a fix in that area, using a slightly different approach (asking because your PR branches off of a commit prior to my fix ^^)

stremlenye:
I'll check tomorrow, but if I remember correctly, the warnings still were there for the widgets with high update frequency.

stremlenye:
I've checked it briefly and can still see the warnings about missing signal handlers on the generated lists updates if using the main branch.

EuanFH:
Thought i would just chime in and say this fixes a similar issue for me.
When i open a window using an onclick event in current master the event will run multiple times.
This branch this fixes this issue for me.
@stremlenye thanks for the fix!

elkowar:
I guess this does seem to fix the issue, nice!
I don't currently like how we're now manually repeating the name of the event handler so often....
The only real alternative I'm able to think of rn is to use some other unique identifier -- there's no inherent benefit to having it be the name of the event handler, so we could technically just use something like `::std::concat!("signal-handler-id:", ::std::line!())` as the key.
That would be unique per macro invocation, and thus unique per event handler... What do you think?

elkowar:
Did that and rebased! Thanks for your work here, I didn't even know about the gtk data thing, and would have taken a while to fully figure this out!

PR #449, opened 2022-04-30
Updated dependencies

pentamassiv:
## Description
16 our of the 68 dependencies are currently outdated. Three of these maybe insecure and could pose a security risk so I updated them and did the necessary changes. I also added a badge to the README so that it is obvious if any dependencies are outdated.

## Additional Notes
I also changed one of the methods of gtk-layer-shell, because the one you were previously using was deprecated with v0_6 of the library. This is the new method:

```
// Sets the keyboard interactivity
gtk_layer_shell::set_keyboard_mode(&window, gtk_layer_shell::KeyboardMode::Exclusive);
```
You might want to check if that is your desired KeyboardMode or not and if you want to make the old method available via a feature flag or something similar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,3 +46,4 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Added more examples
 - List system-level dependencies in documentation
 - Document structure of magic variables (By: legendofmiracles)
+- Updated dependencies
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,9 +95,9 @@ checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
 name = "beef"
-version = "0.5.1"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bed554bd50246729a1ec158d08aa3235d1b69d94ad120ebe187e28894787e736"
+checksum = "3a8241f3ebb85c056b509d4327ad0358fbbba6ffb340bf388f26350aeda225b1"
 
 [[package]]
 name = "bincode"
@@ -167,9 +167,9 @@ checksum = "2fff2a6927b3bb87f9595d67196a70493f627687a71d87a0d692242c33f58c11"
 
 [[package]]
 name = "cfg-expr"
-version = "0.10.2"
+version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5e068cb2806bbc15b439846dc16c5f89f8599f2c3e4d73d4449d38f9b2f0b6c5"
+checksum = "0aacacf4d96c24b2ad6eb8ee6df040e4f27b0d0b39a5710c30091baa830485db"
 dependencies = [
  "smallvec",
 ]
@@ -294,7 +294,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f877be4f7c9f246b183111634f75baa039715e3f46ce860677d3b19a69fb229c"
 dependencies = [
  "quote 1.0.18",
- "syn 1.0.92",
+ "syn 1.0.96",
 ]
 
 [[package]]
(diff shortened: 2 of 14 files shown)


elkowar:
very cool, thanks a lot! Sorry for only getting to this so late ^^

could you rebase your changes onto the latest master, so I can do a final review and merge this?

pentamassiv:
I did the rebase but there are two problems remaining with this PR.

1) The "config::test::test_config" test fails. I always get the following error:

`test config::test::test_config ... FAILED

failures:

---- config::test::test_config stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Differences ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: crates/yuck/src/config/snapshots/yuck__config__test__config.snap
Snapshot: config
Source: crates/yuck/src/config/test.rs:28
-old snapshot
+new results
────────────────────────────────────────────────────────────────────────────────
config.unwrap()
────────────┬───────────────────────────────────────────────────────────────────
   67    67 │         name_span: Span(464, 467, 0),
   68    68 │       )),
   69    69 │       resizable: true,
   70    70 │       backend_options: BackendWindowOptions(
   71       │-        wm_ignore: false,
   72       │-        sticky: true,
   73       │-        window_type: Dock,
   74       │-        struts: StrutDefinition(
   75       │-          side: Left,
   76       │-          dist: Pixels(30),
   77       │-        ),
         71 │+        exclusive: false,
         72 │+        focusable: false,
   78    73 │       ),
   79    74 │     ),
   80    75 │   },
   81    76 │   var_definitions: {
────────────┴───────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
thread 'config::test::test_config' panicked at 'snapshot assertion for 'config`

I don't know what causes it since I have never worked with insta. You probably know better. I would guess that there isn't a problem with my code changes since they only fix building the code? Maybe you can let me know what the issue is and how to fix it?

2) The second problem is the "build" workflow failing. This is a problem with versions of the gtk-layer-shell library. The error is happening because the set_keyboard_mode function is only available in gtk-layer-shell 0.6 and above. Ubuntu is shipping a very old and outdated version of gtk-layer-shell. Jammy (the new LTS version of Ubuntu) has an updated version of it. Github Actions has a public beta of it and I'd expect them to change the Github Actions runner of ubuntu_latest to Jammy later this year so then at least the build error would not be a problem. 
However users might still use old Ubuntu versions or other distros that ship outdated versions ([many of them](https://repology.org/project/gtk-layer-shell/versions) ship 0.6 or later though). They would have to use the deprecated 'set_keyboard_interactivity' function instead. There are multiple options how to handle this and I don't know what you prefer.

1. You could require an updated gtk-layer-shell version and switch the Github Actions runner to Jammy to fix the error
2. You could add the '[system-deps](https://crates.io/crates/system-deps)' crate to the dependencies. It allows you to check the version of the installed library when building the code to find out which function to use.
3. You could add a feature flag for the user to select it manually so no additional dependencies are added

elkowar:
W.r.t. the gtk-layer-shell version -- it sounds like the likelyhood of many users still being stuck on the older version of gtk-layer-shell. In that case, I'd think that it may make sense to wait with updating that specific dependency until the current version is not gonna cause any noticable issues for users -- assuming there's no urgent reason to update (i.e. some weird vulnerability or sth), I'd not see much need to overcomplicate 1. building for some users, 2. the codebase or 3. both ^^

elkowar:
I can't reproduce the failing tests with your PR branch currently, so it's hard to say what's going on there

pentamassiv:
Okay, if you can't reproduce the failing tests, I guess that is a good sign.

Since I am the author of the gtk-layer-shell crate, I disabled setting default features in the crate and pushed a new version. With the new version, there are no problems anymore and it doesn't matter which gtk-layer-shell version the users have. It was a bad idea to set the default features so it was good to work on this PR.

I also ran the Github Actions on my fork and they all passed so I think this should be good to go for your final review :)

pentamassiv:
@elkowar is there anything else you need me to do for you to be able to merge this?

elkowar:
Don't think so, just didn't get around to do a final review and tests yet, sorry ^^'

PR #457, opened 2022-05-14
Add overlay widget

viandoxdev:
## Description

This PR adds the overlay widget, that overlays its children instead of placing them next to each others.

## Usage

```
(overlay
  (child 1)
  (overlayed child)
  ...)
```
The overlay widget takes no props, and takes the size of its first child.

## Checklist


elkowar:
I'm not sure I love that API, but I've also spent two weeks trying to figure out a better one and can't think of anything -- so I'll merge this for now, and worst case if I come up with something I'll make a breaking change ^^ Thanks a lot!

PR #461, opened 2022-05-22
Fix magic var doc being empty

viandoxdev:
## Description

The magic vars documentation is empty (probably since 6b5c35cbf9e921beb68b4a108762775f3674acf5).
This PR allows for `@desc` items to have no matching `@prop`, this should fix the page being empty.

The markdown generation works fine, I haven't built the whole website tho.

### DIFF
--- a/gen-docs.ts
+++ b/gen-docs.ts
@@ -12,24 +12,57 @@ interface Widget {
     isVisible: boolean;
 }
 
+interface MagicVar {
+    name: string;
+    desc: string;
+    prop?: string;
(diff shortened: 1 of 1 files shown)


elkowar:
I'm stupid lmao, thanks!

elkowar:
Should really add CI that builds documentation and shows a diff there,....

viandoxdev:
The github page action isn't running because it only updates when gen-doc.**js** changes, and not **ts**

PR #491, opened 2022-06-26
eww.md updated with void packages

aspeky:
i saw many suffering to find out the packages wish this helps and saves thier time!

## Description

Provide a short description of the changes your PR introduces.
This includes the actual feature you are adding,
as well as any other relevant additions that were necessary to implement your feature.

## Usage

When adding a widget or anything else that affects the configuration,
please provide a minimal example configuration snippet showcasing how to use it and

### Showcase

When adding widgets, please provide screenshots showcasing how your widget looks.
This is not strictly required, but strongly appreciated.

## Additional Notes

Anything else you want to add, such as remaining questions or explanations.

## Checklist

### DIFF
--- a/docs/src/eww.md
+++ b/docs/src/eww.md
@@ -37,6 +37,18 @@ The following list of package names should work for arch linux:
 
 </details>
 
+<details>
+<summary>VoidLinux</summary>
+
+- cairo-devel 
+- pango-devel 
+- glib-devel 
+- atk-devel 
+- gtk+3-devel 
+- gdk-pixbuf-devel
+
+</details>
+
 (Note that you will most likely need the -devel variants of your distro's packages to be able to compile eww.)
 
 ### Building


docs/src/eww.md
```diff
@@ -37,6 +37,18 @@ The following list of package names should work for arch linux:
...
 
+<details>
+<summary>VoidLinux</summary>
+
+- cairo-devel 
+- pango-devel 
+- glib-devel 
+- atk-devel 
+- gtk+3-devel 
+- gdk-pixbuf-devel
```
monke0192, line 48: Only the packages `gtk+3-devel` and `glib-devel` are required. The rest are dependencies of the aforementioned packages

aspeky: still it might give a hand for newbies if they face any issues

monke0192: > still it might give a hand for newbies if they face any issues

They probably won't find any

elkowar:
I'm not a fan of adding specific distro packages do the docs honestly, as that may quickly become a sizeable maintenance burden and as then I'd need to pretty much add docs for every big distro -- I really don't wanna go down that rabbit hole honestly

any alternative ideas for how we may document this a bit better? we could for example do a discussions category or sth where people could write up these things for their specific distros and then link to that, making it a bit more clearly user-submitted, and thus less of a burden on core maintainers ^^'

PR #522, opened 2022-08-11
Allow output names to select a monitor

### LINKED ISSUE #516: [BUG] unpredictable `monitor` values

### Description of the bug

When monitors are dynamically added (as in, docking events), the `monitor` references can change to which monitor they refer. In particular, even when the WM (e.g. bspwm) and X think the monitors are the same configuration, it's a coin toss as to which monitor eww will think is `1`, and which is `2`. This results in bars sometimes swapping monitors on which they appear between dock events.

Of note is that monitor `0` on a laptop (the integrated screen) is always assigned the integrated monitor.

### Reproducing the issue

1. Have a laptop and two external monitors connected to a dock
2. Configure 3 different bars (`bar0`, `bar1`, `bar2`), one on each monitor (laptop gets `:monitor 0`, the others get `:monitor 1` and `:monitor 2`)
3. `eww daemon`
4. `eww open-many bar0 bar1 bar2`
5. Undock the laptop
6. Close the bars, (`eww close-all`)
7. Open one just for the laptop (`eww open bar0`) -- this and step 9 can probably be omitted
8. Dock the laptop
9. Close the one bar (`eww close bar0`)
10. Re-open the bars: `eww open-many bar0 bar1 bar2`

### Expected behaviour

When the WM (bspwm) and X consistently identify the monitors and monitor geometry, eww should also always assign monitors the same identifiers. 

### Additional context

When I dock my laptop, X invariably assigns the left monitor the name `DP2-2` and the right `DP2-1`. bspwm uses the X monitor names, and therefore scripts to put desktops on monitors can consistently lay out workspaces.

A wild guess is that the eww `:monitor` identifiers are simply array indexes, and that the array is built in an order that is not consistent, probably as a result of some underlying system returning an unordered monitor list -- that is, in eww, monitor ordinality is not preserved between monitor connection events. 

A fix would be to use display identifiers instead of indexes. Another fix would be to sort the indexes by display name so that the order is consistent. A third would be to use monitor edid ([eds](https://github.com/compnerd/eds) has some sample code), but edids can be troublesome for a number of reasons.

My preferred solution would be to use display names; X is pretty consistent with assigning consistent display names, and it's a common way to refer to monitors in X programs: `xrandr`, `bspc -M -m --names`, `i3-msg -t get-outputs`, etc. all return the X identifier names, and those names are commonly used in configuring window managers and other X tools.

PillTime:
## Description

This PR proposes to change how monitors are selected, switching from using an index value to the monitor's identifier.
Partially fixes #516.

## Usage

Instead of doing `:monitor 0`, you can simply do `:monitor "DP-1"` for example.

## Additional Notes

So far X11 is supported.
I also changed the code I assume is needed to enable Wayland support, but when testing it, Wayland didn't work.
(I tested Wayland with [Hyprland](https://github.com/hyprwm/Hyprland), using the output of `hyprctl monitors` for the monitor names.
If anyone knows how to fix this, I'd love the help.

- [x] X11
- [ ] ~Wayland~

I've faced the problem of eww bars changing monitor from plugging/unplugging monitors before, so I decided to try and fix this.
~After Wayland support starts working I'll write something in the changelog.~

## Checklist

- [X] I added my changes to CHANGELOG.md, if appropriate.
- [X] The documentation in the `docs/content/main` directory has been adjusted to reflect my changes.


viandoxdev:
`monitor.model()` returns a more complicated string on wayland (at least on sway):
```shell
$ swaymsg -t get_outputs                              
Output eDP-1 'Thermotrex Corporation TL140BDXP01-0 Unknown' (focused)
#             ^        make          ^   model     ^serial
```
does seem to work with it, but this needs to be documented at least.
Might be nice to also have a way to get both behavior for simplicity (and make sure the example works on any hardware)

VuiMuich:
May I suggest to allow both, the numerical ID and the string identifier.
You could add a custom type `MonitorID` that has a tuple `(i32, String)` and a couple of methods de-/serialize and process the tuple data accordingly.
Idk if `eww` does something similar for values that could either be `percentage` or `pixels` too. If not and you would like to have an example of what I am thinking of, I can try to excerpt the relevant code from `leftwm`'s handling of this in `scratchpads` for you.

PillTime:
@viandoxdev I made a small project to test it out and yep, `.model()` just outputs the actual model of the screen in Wayland.
I would prefer to use output names like what `.model()` gives in X11.

I tried to figure out how to get the output name in Wayland, but I've not been successful.
The closest I feel I've gotten was the [`wl_output`](https://gtk-rs.org/gtk3-rs/stable/latest/docs/gdkwayland/struct.WaylandMonitor.html#method.wl_output) name, but that has the format `wl_output@<number>`, so it's not what we want.
<details>
<summary>code from the test project</summary>

```rust
use gdk::{Display, init, prelude::Cast};

fn main() {
    init();
    let display = Display::default().expect("get default display");
    println!("{:?}\n", display.name());
    for m in 0..display.n_monitors() {
        if let Some(mon) = display.monitor(m) {
            if let Some(waymon) = mon.downcast_ref::<gdkwayland::WaylandMonitor>() {
                println!("{:?}", waymon.wl_output());
            } else {
                println!("not wayland monitor");
            }
            println!("{:?}", mon.model());
            println!("{:?}\n", mon.manufacturer());
        } else {
            println!("None\n");
        }
    }
}
```
</details>

Anyway, even if we had to resort to using the screen's model name, if someone has two screens that are the exact same, which I'm sure a lot of people with dual monitor setups have, won't the output of both `.model()` and `.manufacturer()` be the same?

---

@VuiMuich I'm all for giving users more options, and allowing both numbers and names makes this change non-breaking.
I think the best would be an `Enum` with a variant that stores an `i32` and another that stores a `String`.
I'll see what I can do, but it's probably not gonna be today.

viandoxdev:
Regarding the wl_output, what you got isn't the name, but probably just a debug representation of the handle used by wayland-client for the underlying wl_output, so a clearer name might be accessible, but the current api of wayland-client (the latest stable version) doesn't seem to allow for that. The beta might be able to do it but since gtk rs relies on the "older" version its not really possible

elkowar:
I do like having the option to provide the monitor name, definitely!  I'm not sure about the wayland side of things here either, but the X API looks good. although I agree with @viandoxdev in that it'd be good to support both, as hard coding a monitor identifier will make widgets suddenly be way more setup specific and thus no longer a download->run->done for people just trying out widgets other people made. keeping the portability and backwards compatibility here would be great

VuiMuich:
Oh, right `enum` makes much more sense and is what `leftwm` [has as well](https://github.com/leftwm/leftwm/blob/b2e98600e33f8ace715aff12bc659eb6a7f23297/leftwm-core/src/models/size.rs).

viandoxdev:
Last update for the wayland monitor names, the connector (i.e "DP-1", "HDMI-2") *is* known by gtk since 3.24.10, although the gtk-rs api doesn't provide a way to query them. They can still be accessed through some questionable unsafe code, but because it wouldn't even work on older versions of gtk we can't really use them for monitor selection. The only real way to use them would be to match the monitor string against both the model and the connector (if available) to let people use what they want, but at this point I don't know if its really worth it (would need to be documented).

[code to get the connector](https://gist.github.com/viandoxdev/15a85ecdaa22e088f51441ff2b591073)

PillTime:
I created an enum for monitor identifiers. Both `eww` and `yuck` need access to it, so I decided to put it in `eww_shared_util`.
It has a method to return `Option<Monitor>` so that the search loop is only written in one place.

@viandoxdev Nice work figuring out how to get the connector in wayland. I feel like the best for now would be to only allow connector names in X11, and use just the monitor numbers in Wayland. Whenever gtk provides a way to query connectors, we can simply update the code to use it.

@elkowar I'm ok with showing the available monitors on error.

xxxserxxx:
This is a thing of beauty. It brings a tear to my eye.

Works wonderfly, thank you!

PillTime:
Changed so that when using Wayland, eww only allows numeric identifiers, and will give an error if the user tries to use names.

elkowar:
Could you allow edits from maintainers? I did some small cleanups to the code, once those are done I'd be down to merge this

PillTime:
Edits were already enabled, but i disabled and reenabled them just in case that option was bugged.

elkowar:
Thanks a lot for your work!

PR #557, opened 2022-09-08
Fix flake refactor ([#532])

PhilTaken:
## Description

The exposed overlay did not work as a standalone, required `final` to be overlayed with rust-overlay prior.
This was caused by the `mkRustToolchain` function and the way it was used in the overlay itself.

To solve this, I extended the passed pkgs set with the rust-overlay overlay first, to gain access 
to the rust-bin set from arbitrary pkgs passed into the `mkRustToolchain` function.

## Additional Notes

I also added some tests that depend on the packages definition, which only uses the overlay 
exposed by self, which in turn defines the build process for any pkgs passed in.

## Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,7 +10,6 @@
     let
       pkgsFor = system: import nixpkgs {
         inherit system;
-
         overlays = [
           self.overlays.default
           rust-overlay.overlays.default
@@ -18,7 +17,12 @@
       };
 
       targetSystems = [ "aarch64-linux" "x86_64-linux" ];
-      mkRustToolchain = pkgs: pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+      mkRustToolchain = pkgs: let
+        inherit (pkgs.lib) fix extends;
+        rpkgs = if pkgs.lib.hasAttrByPath [ "rust-bin" ] pkgs
+          then pkgs
+          else fix (extends (import rust-overlay) (self: pkgs));
+      in rpkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
     in
     {
       overlays.default = final: prev:
@@ -41,14 +45,10 @@
           eww-wayland = final.eww.override { withWayland = true; };
         };
 
-      packages = nixpkgs.lib.genAttrs targetSystems (system:
-        let
-          pkgs = pkgsFor system;
-        in
-        (self.overlays.default pkgs pkgs) // {
-          default = self.packages.${system}.eww;
-        }
-      );
+      packages = nixpkgs.lib.genAttrs targetSystems (system: {
+        inherit (pkgsFor system) eww eww-wayland;
+        default = self.packages.${system}.eww;
+      });
 
       devShells = nixpkgs.lib.genAttrs targetSystems (system:
         let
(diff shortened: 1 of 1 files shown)


PhilTaken:
@fortuneteller2k Did you check if the overlay works?

fortuneteller2k:
its intentional for users to import their own rust-overlay

PhilTaken:
> its intentional for users to import their own rust-overlay, it allows people to keep up with rust-toolchain.toml changes when the lockfile isnt updated

that seems very counter-intuitive, the end-user should not have to deal with that.
lockfile changes should be dealt with on the repository side, immediately

PhilTaken:
@fortuneteller2k if need be the version of rust-overlay can still be overriden via `inputs.nix-overlay.follows`

flake.nix
```diff
@@ -10,15 +10,13 @@
...
-          self.overlays.default
-          rust-overlay.overlays.default
-        ];
+        overlays = [ self.overlays.default ];
       };
 
       targetSystems = [ "aarch64-linux" "x86_64-linux" ];
-      mkRustToolchain = pkgs: pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+      mkRustToolchain = pkgs: let
+        rpkgs = pkgs.extend rust-overlay.overlays.default;
```
fortuneteller2k, line 18:
```suggestion
```

I would rather pass `pkgs.extend rust-overlay.overlays.default` where necessary, like in line 24, it's an expensive operation that causes a full rebuild of nixpkgs.

For `packages` and `devShell` `rust-overlay.overlays.default` should be initialized in `pkgsFor`

PhilTaken: what about `lib.fix (lib.extends (import sources.rust-overlay) (self: pkgs))`. that should have less overhead than `pkgs.extend` according to https://github.com/oxalica/rust-overlay/issues/94.

fortuneteller2k: something the likes of this will work

```diff
diff --git a/flake.nix b/flake.nix
index d097383..98a8040 100644
--- a/flake.nix
+++ b/flake.nix
@@ -23,7 +23,10 @@
     {
       overlays.default = final: prev:
         let
-          rust = mkRustToolchain final;
+          rust =
+            if nixpkgs.lib.hasAttrByPath [ "rust-bin" ] final
+            then mkRustToolchain final
+            else mkRustToolchain (prev.extend rust-overlay.overlays.default);

           rustPlatform = prev.makeRustPlatform {
             cargo = rust;
```

flake.nix
```diff
@@ -10,15 +10,16 @@
     let
       pkgsFor = system: import nixpkgs {
         inherit system;
-
-        overlays = [
-          self.overlays.default
-          rust-overlay.overlays.default
-        ];
```
fortuneteller2k, line 17: revert this and after it lgtm

PhilTaken: done

PhilTaken:
@elkowar is there anything blocking a merge?

w-lfchen:
Can this pr be closed?
I don't think it's needed anymore

elkowar:
Closing as I think this is no longer necessary, but given that I have no idea about nix stuff, feel free to re-open or create a new PR if issues still come up! sorry for the delay

PR #577, opened 2022-09-19
Safe Access Operator

### LINKED ISSUE #565: [FEATURE] Safe Access Syntax

When writing my configuration files, I often find myself writing something like

```
(some-widget
  :some-arg { global-state["optional-value"] != "" ? "prefix${global-state["optional-value"]}" : "" })
```

Seeing as we already have the Elvis operator `:?` (with the `""` value being considered equivalent to `None`), this proposes adding a new syntax for doing the opposite.

The syntax for this, heavily inspired by Kotlin's [safe access](https://kotlinlang.org/docs/null-safety.html#safe-calls) operator.

The two main syntaxual forms are as follows

```
{ global-state["optional-value"]?.let { gval -> "prefix${gval}" }

{ global-state["optional-value"]?.let { "prefix${it}" }
```

Where in the second, we elide the binding identifier, opting to use the implicit `it`. In kotlin, it is standard practice to outlaw name shadowing (including the implicit `it`), but I don't know that that would be strictly necessary.

In kotlin this functionality is actually the confluence of 3 separate things:

- Safe Call/Access
- Extension functions (`let` is a generic inline extension on all types)
- Trailing Lambda Syntax (no parens around `{ gval -> ... }`)

I am not suggesting implementing all of the above, as that would be too heavy for simplexpr, but instead just this specific usage as it would round out the elvis operator and greatly simplify writing configurations.

### Additional context

_No response_

oldwomanjosiah:
## Description

- closes #565
- Updates the elvis operator `?:` to treat a JSON `Null` the same as the empty string `""`  
  This was done, as that is what is returned by an object access for an undefined field.

## Usage

```
:label { labels?.workspaces?.one ?: "Workspace One" }
```

with some varable `labels`. `labels.workspaces.one ?: "Workspace One"` would error if:

- `labels` was not an object
- `labels.workspaces` was not an object

It would also act strangely if either any of `labels`, `.workspaces`, or `.one` were `Null`, as the elvis operator treated that as "already existing".

[Diffs containing full example of usage](https://github.com/oldwomanjosiah/dots/compare/220108101c45d10556dd110b73585b734ead72a0...dc10ecddda61503ec479252a218cf40e40ba1a6c)

## Additional Notes

It might also make sense to allow for `opt-object?.[key-expr]`, but that is left for a follow-up.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Add support for output names in X11 to select `:monitor`.
 - Add support for `:active`-pseudoselector on eventbox (By: viandoxdev)
 - Add support for `:password` on input (By: viandoxdev)
+- Add support for safe access (`?.`) in simplexpr
 
 ### Notable fixes and other changes
 - Scale now only runs the onchange command on changes caused by user-interaction
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to eww will be listed here, starting at changes since versio
 
 ## [Unreleased]
 
+### Features
+- Add support for safe access (`?.`) in simplexpr (By: oldwomanjosiah)
+
 ## [0.4.0] (04.09.2022)
 
 ### BREAKING CHANGES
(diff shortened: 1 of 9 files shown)


elkowar:
thanks a lot for your work! This is cool!

Issue #408, opened 2022-02-24
[FEATURE] Let-in syntax in simplexpr


elkowar:
### Description of the requested feature

The more complexity simplexpr allows for, the more use there would be for variables within expressions.
Given #324, it might make sense to try and come up with some unified syntax or principle for this, although having separate syntax in simplexpr _might_ not be an issue, either.
Alternatively, it could be stated that it'd make sense to just rely on the local variables requested in #324, and have separate expressions.

However, I could still see having let-in syntax for simplexpr explicitly be a worthwhile addition.

### Proposed configuration syntax

```
let foo = "hi"; bar = "ho"; in foo + bar

let
  foo = "hi"
  bar = "ho"
in
  foo + bar
```

(not sure about newline sensitive stuff here, not really a fan :/

### Additional context

One alternative to this has been proposed with the `with` function in #407

MartinJM:
The `with` function from #407 works as follows:

- `with(name, value, expression)`: Evaluate the simple expression (which must be a string), where the variable `name` has value `value`

Example:

```
with('a', '{"0": "hello ", "1": "world"}', 'a[0] + a[1]')
hello world
```

That implementation can be found here: https://github.com/MartinJM/eww/commit/a4e2095a8a6d50dab28cd69681fe2eb818e15226

---

I do have a question about the let-in syntax: What would the scope of the variable be? The first expression after the `in`, or all expressions until the end?

If I may also add my own opinions:

If (Simplexpr) variables are not supported in local variables, I think it would make sense to have a different syntax. If the syntax is the same, I think it will lead people to believe that the functionality is the same as well.

I'm personally also not really a fan of newline-based stuff. But besides that, I think that newlines in a Simplexpr aren't really supported at the moment? At least, I haven't found a way to have them in a Simplexpr.

oldwomanjosiah:
Would it make sense to instead follow what ml does here with

```
let
  ; reserved keywords let you know where the next def starts, and what part is a name
  val a = someValDef
  val b = someValDefAgain
in
  ; some expr using a & b
end ; clear closing here
```

viandoxdev:
Is the `in` really necessary ? Can't we just have the variables be implicitly scoped to the whole expression ?
Also for the newlines, maybe just add a end delimiter like a coma ?
```
let a = valOrExpr, a + 1 ; Or any other expression that uses a
```

oldwomanjosiah:
[...]

PR #578, opened 2022-09-19
Add Let In Syntax

oldwomanjosiah:

## Description

This is not intended to be final, but I had some time so I thought I would work on this. First pass ended up being easier than I was expecting.

- Implements #408 

## Usage

```
(defvar somevar {{ "key": "Some Value" }})
(label
  :text { let it = somevar.value in "Here Is ${it}" end }
  )
```

### Showcase

When adding widgets, please provide screenshots showcasing how your widget looks.
This is not strictly required, but strongly appreciated.

## Additional Notes

I borrowed the testing macro from #577

## Checklist

### DIFF
--- a/crates/simplexpr/src/eval.rs
+++ b/crates/simplexpr/src/eval.rs
@@ -1,7 +1,7 @@
 use itertools::Itertools;
 
 use crate::{
-    ast::{BinOp, SimplExpr, UnaryOp},
+    ast::{AccessType, BinOp, DefinitionList, SimplExpr, UnaryOp},
     dynval::{ConversionError, DynVal},
 };
 use eww_shared_util::{Span, Spanned, VarName};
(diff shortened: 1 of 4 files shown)


crates/simplexpr/src/eval.rs
```diff
@@ -91,6 +91,9 @@ impl SimplExpr {
                     .collect::<Result<_, _>>()?,
             ),
             x @ Literal(..) => x,
+            LetIn(_, d, b) => {
+                todo!("Resolve Refs is unused, TODO(josiah)");
```
oldwomanjosiah, line 95: It's unclear to me atm what should be done here. As far as I can tell the function is (transitively) unused, and could be removed, so it might not be a problem.

It might make the most sense to use the values from the let block to map the inner vars, but the variable defs haven't been evaluated yet here.

oldwomanjosiah:
@elkowar Just following up here to see if you have any feedback on the direction of this? Specifically if `try_map_var_refs` can/should be removed, or if not what the behavior there should be, but also in general.

PR #590, opened 2022-10-04
Add Test runs to CI, don't run `gh-pages` from forks

oldwomanjosiah:
Skipping the usual PR template, since this is just CI

### Changes

- Use matrix for builds, so that they can be done in parallel
- Add `test` workflow, using nextest to run all tests in repo
- Update formatting in all workflow files to follow yaml best practice
- Skip documentation workflow when dispatched on a fork


oldwomanjosiah:
It might be worth it, to lower check times, to run the CI in a container image with the deps already installed. 

https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

crates/eww/src/main.rs
```diff
@@ -190,3 +190,13 @@ fn check_server_running(socket_path: impl AsRef<Path>) -> bool {
...
+
+
+
+
+
+
+
+
+
+
```
w-lfchen, line 202: is this a feature? ^^

oldwomanjosiah:
This is pretty old, and definitely outdated.

w-lfchen:
i'll see whether i can pick up the ideas of this pr at some point if that's fine ^^

oldwomanjosiah:
Yeah for sure! I don't run Linux as my daily right now so I can't really contribute there, but I'd love to see it get in fs

PR #595, opened 2022-10-10
docs: Add example for data structures

PrettyCoffee:
## Description

In my opinion the repo was missing an example on how to create and use more advanced data types than just number arrays. I had to read through issues and other widget implementations to create this example and felt like the documentation would benefit from it :)

## Additional Notes

I wanted to create the most basic example I could come up with. If you dislike the theme or want something more practical / have an idea what I could put there instead while keeping it simple, let me know :)

Furthermore I think the documentation would overall benefit from more small and simple examples that explain single mechanics, just as a feedback note. The documentation that exists is easy to follow and has a very good structure! It just needs a little more depth.

## Checklist

-> seem not to apply to documentation-only changes


elkowar:
~~Only 2 years later:~~ Thanks a lot for your work! This is definitely documentation that should have been there from the start.

PR #609, opened 2022-10-29
Fixes #608 make exclusive more configurable on wayland

### LINKED ISSUE #608: [FEATURE] Overlap waybar's exclusive zone.

I know this is completely useless for 99% of the users, but I think it would be cool if eww widgets could invade waybar's reserved space. I'm trying to make a fullscreen eww widget and waybar is in the way. Same thing happens when I set a eww bar and make it exclusive. If I were in X I guess I could just fullscreen eww with wm because it can be, but on wayland I can't since it's not a window.

_No response_

### Additional context

_No response_

viandoxdev:
## Description

Changes the exclusive property on wayland to allow for another option: ignore

## Usage

```clojure
  :exclusive "exclusive" ; reserve space
  :exclusive "normal" ; default behavior
  :exclusive "ignore" ; don't move window if occluding another
```

## Additional Notes

Not a fan of the names `exclusive`, `normal`, `ignore`, `:exclusive "exclusive"` looks weird and they're not really explicit.
Also I might have broken a snapshot test, cargo test didn't fail tho.

## Checklist

### DIFF
--- a/crates/eww/src/display_backend.rs
+++ b/crates/eww/src/display_backend.rs
@@ -14,6 +14,7 @@ mod platform {
     use gdk;
     use gtk::prelude::*;
     use yuck::config::{
+        backend_window_options::ExclusiveZone,
         window_definition::{WindowDefinition, WindowStacking},
         window_geometry::AnchorAlignment,
     };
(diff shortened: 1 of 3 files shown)


elkowar:
Yea, having the breaking change here is kind of annoying... I'd love to either include this in a release that already includes other breaking changes, or make it _not_ break the current format. Any ideas as to how you could implement this without breaking backwards-compatibility?

w-lfchen (commented):
i propose still supporting `exclusive` and adding a new option instead, maybe something like `reserve`, in which the user may directly define the amount of space to reserve, that number is then passed to `gtk_layer_shell::set_exclusive_zone` directly, overriding `exclusive`.
the `exclusive` docs could be updated to mention that this simply automatically reserves space.

this would allow the use case of someone wanting to partially overlap their bar over windows if i understand [the protocol](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:set_exclusive_zone) right.
don't know whether someone would do that, but providing more options hardly seems like a bad idea.

zane-weissman:
What about an option like `:wl-ignore`? For X11 there is `:wm-ignore` which does something similar. If both `:wl-ignore` and `:exclusive` are set to true, one would have to be chosen over the other because a window cannot have both properties, but it would keep the option values simple (just true/false) and avoid breaking backwards-compatibility.

PR #629, opened 2022-11-27
Support empty string for safe access operator Currently only an empty json string, i.e. a string in a string was supported: `'""'`.

ModProg:

## Description

Fixes `?.` for empty strings

## Usage

{""?.test == "null"}

## Additional Notes

TBD: should a json string i.e. `'""'` be considered `null`?

## Checklist

### DIFF
--- a/crates/simplexpr/src/eval.rs
+++ b/crates/simplexpr/src/eval.rs
@@ -232,24 +232,31 @@ impl SimplExpr {
 
                 let is_safe = *safe == AccessType::Safe;
 
-                match val.as_json_value()? {
-                    serde_json::Value::Array(val) => {
-                        let index = index.as_i32()?;
-                        let indexed_value = val.get(index as usize).unwrap_or(&serde_json::Value::Null);
-                        Ok(DynVal::from(indexed_value).at(*span))
+                // Needs to be done first as `as_json_value` fails on empty string
+                if is_safe && val.as_string()?.is_empty() {
+                    Ok(DynVal::from(&serde_json::Value::Null).at(*span))
+                } else {
+                    match val.as_json_value()? {
+                        serde_json::Value::Array(val) => {
+                            let index = index.as_i32()?;
+                            let indexed_value = val.get(index as usize).unwrap_or(&serde_json::Value::Null);
+                            Ok(DynVal::from(indexed_value).at(*span))
+                        }
+                        serde_json::Value::Object(val) => {
+                            let indexed_value = val
+                                .get(&index.as_string()?)
+                                .or_else(|| val.get(&index.as_i32().ok()?.to_string()))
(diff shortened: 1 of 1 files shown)


crates/simplexpr/src/eval.rs
```diff
@@ -232,24 +232,31 @@ impl SimplExpr {
...
-                    serde_json::Value::String(val) if val.is_empty() && is_safe => {
-                        Ok(DynVal::from(&serde_json::Value::Null).at(*span))
+                // Needs to be done first as `as_json_value` fails on empty string
+                if is_safe && val.as_string()?.is_empty() {
+                    Ok(DynVal::from(&serde_json::Value::Null).at(*span))
+                } else {
+                    match val.as_json_value()? {
+                        serde_json::Value::Array(val) => {
+                            let index = index.as_i32()?;
+                            let indexed_value = val.get(index as usize).unwrap_or(&serde_json::Value::Null);
```
oldwomanjosiah, line 242: This isn't for this PR, but should we be throwing an error here? It doesn't seem like we'd want to silently ignore out of bounds accesses

ModProg:
> What was the error you were seeing when indexing an empty string?

@oldwomanjosiah 

When I add the test to master I get:
```
thread 'eval::tests::safe_access_to_empty' panicked at 'Output was not Ok(_): ConversionError(ConversionError { value: "", target_type: "json-value", source: Some(Error("EOF while parsing a value", line: 1, column: 0)) })', crates/simplexpr/src/eval.rs:392:5
```

elkowar:
I'm not sure if this is actually desirable behavior, as empty strings will be valid values in many situations -- having a differentiation between empty strings and null does seem quite useful. However, I guess it also depends on how empty strings are treated elsewhere, which to be honest, I'm not 100% sure at this point -- and it miiiiiight be useful to then provide some easier way to convert empty strings into null... easier than `(foo == "") ? "null" : foo`

ModProg:
> I'm not sure if this is actually desirable behavior, as empty strings will be valid values in many situations

yes, but `""?.test` would be an error otherwise so this is not breaking anything.

the only situation where behavior wouldn't be unambiguous would be `""?:`.

ModProg:
I think we should only make an empty string `""` and not an empty json string `'""'` null.

If you use a `?` operator you actively decide to use the value as json, so as long as it is documented it is fine to add this behavior.

elkowar:
That i'd be more fine with, sure! Although this behavior should probably be documented _very_ well

w-lfchen (changes requested):
i've adopted this pr in #1176 (i want to finally get this merged), this pr can therefore be closed.
thank you for your work!

elkowar:
Closing, as this has been merged as part of #1176. Thanks a ton for your work!

PR #640, opened 2022-12-03
Make onclick trigger for both mouse and keyboard

julianschuler:
## Description

Currently, the `onclick` command for buttons is only executed when the button is clicked using the left mouse button. This pull request changes the behavior of the `onclick` handler to also be executed if the button is activated using the keyboard (with enter or space).

## Usage

Here is a simple example button:

```
(defwindow example
  :monitor 0
  :focusable true
  (button :onclick "notify-send 'Hello'"
    "Test button")
)
```

Add the snippet to your `eww.yuck` and open the widget using

```
eww open example
```

You will see a notification for both a left click as well as enter and space.

## Additional Notes

Currently, the `onclick` command is activated as soon as the button is pressed. A side effect of the new implementation is that the command is now executed after the button is released.

Furthermore I was not able to find an equivalent signal handler for the event box widget, hence the keyboard support is only present for the button widget.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 ### Features
 - Add support for safe access (`?.`) in simplexpr (By: oldwomanjosiah)
 - Allow floating-point numbers in percentages for window-geometry
+- Make button onclick handler trigger for both mouse and keyboard
 
 ## [0.4.0] (04.09.2022)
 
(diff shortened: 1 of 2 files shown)


w-lfchen (changes requested):
this code is needlessly complicated.

also, there have been changes here due to the gtk updates in #1144, be mindful of that if you decide to fix this pr.

w-lfchen:
#1187 supersedes this pr which can therefore be closed.
you have been attributed, thank you for your work!

w-lfchen (changes requested):
#1187 supersedes this pr which can therefore be closed.
you have been attributed, thank you for your work!

PR #645, opened 2022-12-13
add `EWW_BATTERY` support in (free|open|net)bsd

dangerdyke:
## Description

This should add battery info support for 3/4 of the *BSDs supported as rust compile targets. I have only tested on FreeBSD, but based on the man pages of the others it should work there as well.

## Usage

unchanged

### Showcase

```
$ target/release/eww get EWW_BATTERY
{"BAT0": { "status": "high", "capacity": 61 }, "total_avg": 61}
$ uname -o
FreeBSD
```

## Additional Notes

I was not able to add support for dragonflyBSD, as it's version of the `apm` command is aliased to `acpiconf -i 0`, which has different output and only lists the first battery. The `htop` source code seems to implement this for dragonfly using something to the effect of `sysctl -n hw.acpi.battery.life`, but I could not find documentation for that sysctl variable anywhere, and don't have a dragonfly install to test on.

## Checklist


crates/eww/src/config/system_stats.rs
```diff
@@ -153,8 +153,54 @@ pub fn get_battery_capacity() -> Result<String> {
...
+    // https://man.netbsd.org/apm.8
+    let mut json = String::from('{');
+    let re_total = regex!(r"(?m)^Remaining battery life: (\d+)%");
+    let re_single = regex!(r"(?sm)^Battery (\d+):.*?Status: (\w+).*?(\d+)%");
+    for bat in re_single.captures_iter(&batteries) {
+        json.push_str(&format!(
+            r#""BAT{}": {{ "status": "{}", "capacity": {} }}, "#,
+            bat.get(1).unwrap().as_str(),
+            bat.get(2).unwrap().as_str(),
+            bat.get(3).unwrap().as_str(),
```
legendofmiracles, line 191: Haven't programmed in months, but I wrote magic vars.
Maybe better to use ? instead of unwrap. Magic variables are a extra, and if they break, they shouldn't take down all of eww.
And it sounds like this is going to break on a lot of bsd distributions (which is fine! Something is better than nothing here)

elkowar: Should be fine in this case, given that regex should guarantee that there will always be three captures,  when the regex pattern has three capture groups, no? Or am I missing something...

elkowar:
Could you rebase this onto latest master? I would do it, and according to github I _should_ be able to, but it seems as though currently I can't push to your branch ^^'
Looking good though, thanks!

dangerdyke:
done!

PR #721, opened 2023-03-25
feat: `truncate-left` for labels

kawaki-san:
## Description

Adds a flag that allows truncation to the left side of text

## Usage

```yuck
(defwidget center []
  (label :text "ラウトは難しいです"
         :truncate-left true
         :limit-width 2
         :show-truncated true
  )
)
```

### Showcase

The configuration above will result in something like:

Likewise, using `false` on the `truncate-left` property or omitting the property completely will default to right hand side truncation.

`:truncate-left true :limit-width 4 :show-truncated false`:

`:truncate-left false :limit-width 5 :show-truncated true`:

`:limit-width 5 :show-truncated true`:

## Additional Notes

It might also be useful to allow users to specify some text to use instead of the ellipsis?

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 ## [Unreleased]
 
 ### Features
+- Add `truncate-left` property on `label` widgets (By: kawaki-san)
 - Add support for safe access (`?.`) in simplexpr (By: oldwomanjosiah)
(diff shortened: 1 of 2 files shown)


elkowar:
Thanks a lot, never really thought about this, but it should definitely exist!

mantacid:
I really wish the docs mentioned this property. I didn’t know about this and had to do something weird with regular expressions.

elkowar:
@mantacid The docs should now be fixed!

PR #747, opened 2023-04-14
Update the gtk libraries to 0.17

ettancos:
Updated the gtk related libraries to ~v0.17

This opens the path to use the updated layer-shell protocol to be able to specify keyboard interactivity to be exclusive or on-demand allowing different use-cases on wayland.

Notable changes:
* v3_22 api is now default
* glib &Self::Type parameters got removed, object is now in self.obj()
* glib properties moved to Builders instead of using `::new` which is now deprecated, I've used the added helper macros to derive them
* `window::set_wmclass` has been removed, according to the docs its redundant so I just removed it

tested on wayland

Maybe some feedback from X users would be welcome just in case.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,9 +144,9 @@ checksum = "2ce4f10ea3abcd6617873bae9f91d1c5332b4a778bd9ce34d0cd517474c1de82"
 
 [[package]]
 name = "atk"
-version = "0.15.1"
+version = "0.17.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


VuiMuich:
So far running well on X11 for me.

elkowar:
This looks really good! Sorry for the long wait ._.

The only thing I'm not super happy with is removing the wm-class -- while GTK docs rightfully say that it's not a good idea to screw with it in regular applications, for eww (where every widget might "mean" something completely different, even though it's the same application) it _does_ provide lots of value to have each widget be clearly identifiable by the WM and related tooling through the WM class

PR #783, opened 2023-05-25
Implement explicitly setting a center of rotation

mario-kr:
## Description

The transform-Widget provides "rotate" to rotate its child. However, the default center of rotation is (0, 0) (aka top-left), so it was not possible to, for example, rotate a child in-place.

This commit implements the additional options `rotation-center-x` and `rotation-center-y`. Both are optional, and the default value is 0.0 for each, so previous configurations should produce the same results.

## Usage

```yuck
(defwindow grid
  :stacking "fg"
  :windowtype "dock"
  :wm-ignore true
  :geometry (geometry
    :x "10px"
    :y "10px"
    :width "100px"
    :height "100px")
  (transform
    ;:rotate {-0.5 * percentloop}
    :rotate 15
    :transform-origin-x "50%"
    :transform-origin-y "50%"
    (box :orientation 'v' :style 'padding: 5px;'
      (box :orientation 'h'
        (label :style 'background-color: grey;' ' ')
        (label :style 'background-color: green;' ' '))
      (box :orientation 'h'
        (label :style 'background-color: red;' ' ')
        (transform
          :rotate {2*percentloop}
          :transform-origin-x "30%"
          :transform-origin-y "50%"
          :translate-x "-20%"
          (box
            (box :space-evenly true
              (label :style 'background-color: yellow;' ' ')
              (label :style 'background-color: black;' ' ')
              (label :style 'background-color: yellow;' ' '))))))))

(deflisten percentloop :initial 0 `i=0; while true; do let i=$i+1; echo $i; sleep .05; done`)
```

```scss
* {
    all: unset;
    margin: 0px;
    padding: 0px;
}
```

### Showcase

I made a 5 second video of above configuration, but can't embed it here.

## Additional Notes

* There are artifacts, if, for example in the above configuration, the transform of the whole grid would be a static rotate. As this does not happen when everything is moving, maybe a redraw is missing somewhere.
* The behaviour if both `rotation-center-*` and `translate-*` needs to be mentioned, though it is not "wrong". If an in-place/centered rotation is wanted, *and* a shift to the right, the shift to the right needs to be added on top of the rotation-center-value to get an in-place rotation.

## Checklist

### DIFF
--- a/crates/eww/src/widgets/transform.rs
+++ b/crates/eww/src/widgets/transform.rs
@@ -173,10 +173,10 @@ impl WidgetImpl for TransformPriv {
                 None => 1.0,
             };
 
-            cr.scale(scale_x, scale_y);
             cr.translate(transform_origin_x, transform_origin_y);
             cr.rotate(perc_to_rad(rotate));
             cr.translate(translate_x - transform_origin_x, translate_y - transform_origin_y);
+            cr.scale(scale_x, scale_y);
 
             // Children widget
             if let Some(child) = &*self.content.borrow() {
--- a/crates/eww/src/widgets/transform.rs
+++ b/crates/eww/src/widgets/transform.rs
@@ -17,6 +17,12 @@ pub struct TransformPriv {
     #[property(get, set, nick = "Rotate", blurb = "The Rotation", minimum = f64::MIN, maximum = f64::MAX, default = 0f64)]
     rotate: RefCell<f64>,
 
+    #[property(get, set, nick = "Transform-Origin X", blurb = "X coordinate (%/px) for the Transform-Origin", default = None)]
+    transform_origin_x: RefCell<Option<String>>,
+
+    #[property(get, set, nick = "Transform-Origin Y", blurb = "Y coordinate (%/px) for the Transform-Origin", default = None)]
+    transform_origin_y: RefCell<Option<String>>,
+
     #[property(get, set, nick = "Translate x", blurb = "The X Translation", default = None)]
     translate_x: RefCell<Option<String>>,
 
@@ -37,6 +43,8 @@ impl Default for TransformPriv {
     fn default() -> Self {
         TransformPriv {
             rotate: RefCell::new(0.0),
+            transform_origin_x: RefCell::new(None),
+            transform_origin_y: RefCell::new(None),
             translate_x: RefCell::new(None),
             translate_y: RefCell::new(None),
             scale_x: RefCell::new(None),
@@ -55,6 +63,14 @@ impl ObjectImpl for TransformPriv {
         match pspec.name() {
             "rotate" => {
                 self.rotate.replace(value.get().unwrap());
+                self.obj().queue_draw(); // Queue a draw call with the updated value
+            }
+            "transform-origin-x" => {
+                self.transform_origin_x.replace(value.get().unwrap());
+                self.obj().queue_draw(); // Queue a draw call with the updated value
+            }
+            "transform-origin-y" => {
+                self.transform_origin_y.replace(value.get().unwrap());
                 self.obj().queue_draw(); // Queue a draw call with the updated value
             }
             "translate-x" => {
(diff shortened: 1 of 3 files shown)


elkowar:
- I feel like we _would_ want the origin to be relative to the position of the widget, such that you _don't_ need to add the translation to the rotation-center value -- especially given that you'll most likely want to use percentages for the origin
- I'd just call it `transform-origin`, keeping the consistency with how the same thing is called in CSS
- I'm not sure if I correctly understand the kind of artifacts you're referring to -- I'll see if I can reproduce them, but obviously optimally, there wouldn't be any visual artifacts here 😉

mario-kr:
Hi, sorry for the late response; I'll try to answer point by point:

* If I understand correctly, then the translate-x/-y values should be added onto the origin, so that e.g. 50%/50% origin will always turn out to be an in-place rotation? As for units, the current implementation understands both percentages and pixels, so maximum flexibility for the user.
* I renamed the variables to `transform-origin`. I am a bit unsure, if this is the same as in CSS, though that may be based on my inexperience with CSS in general.
* IIRC I only changed the first :rotate value to 95 or something similar.

According to the documentation, the current transform-order is: rotate -> translate -> scale. But, looking at the code (https://github.com/elkowar/eww/blob/master/crates/eww/src/widgets/transform.rs#L165) it seems to be: scale -> rotate -> translate
Point 1 might affect this order too; scale after all changes the size of a widget, so any other transformation might possibly be affected. On the other hand, the current implementation retrieves the allocated_width/height exactly once, and uses that same value for all transformations, so the transformation might currently actually be independent of each other (somewhat)?

mario-kr:
Update:
I noticed #1103 when I looked at this project regarding the wrong EWW_NET values, and decided to renew this PR a bit.

- rebased to master, fixed any hiccups due to a different way of defining properties since I last worked on this
- updated the example rotation config

crates/eww/src/widgets/transform.rs
```diff
@@ -150,8 +175,9 @@ impl WidgetImpl for TransformPriv {
             };
 
             cr.scale(scale_x, scale_y);
+            cr.translate(transform_origin_x, transform_origin_y);
             cr.rotate(perc_to_rad(rotate));
-            cr.translate(translate_x, translate_y);
+            cr.translate(translate_x - transform_origin_x, translate_y - transform_origin_y);
```
w-lfchen, line 180: this doesn't break the order defined [here](https://github.com/elkowar/eww/blob/057297b4a339a4f72a6adf4ab587ebde2c331fcb/crates/eww/src/widgets/widget_definitions.rs#L1185), right?

w-lfchen (approved):
lgtm!
just one minor question, see the comment

crates/eww/src/widgets/transform.rs
mario-kr: Well, even before this patch, the methods called on `cr` are in the order:

1. scale
2. rotate
3. translate

whereas in the comment you linked, it's claimed that the order is:

1. rotate
2. translate
3. scale

Sooooo... I don't know? Was it ever in the right order? Or is it sorted by the cairo-Context?

w-lfchen: it doesn't sort them, if i'm understanding [the docs](https://www.cairographics.org/manual/cairo-Transformations.html) right.
i suggest sorting them in the order described in the documentation, though i don't think it makes much of a difference.

also, the only thing causing the merge conflict is the changelog entry, that should be trivial to resolve when rebasing onto master ^^

mario-kr: Screenshot from 2024-08-29 17-46-53

There is a difference I'm afraid.
On the left side, scale is called after rotate + translate. basically, it warps the contents into a parallelogram when rotated. (the scaling is applied in the original directions of the x and y axis)
On the right is the same eww.yuck, but scale is called before rotate+translate, or the whole, scaled widget is rotated.

Generally speaking, users can always force a specific order by nesting multiple transforms. I did this here, where i nested the rotate+translate transform inside the scale transform:
Screenshot from 2024-08-29 18-00-03
but it's not entirely the same; the bottom seems to be cut off (Because the size on the y-axis for this widget was limited by the scale to 60%)

soo... do we want the warping rotate as the default (scale after rotate, as documented), or do we want the non-warping rotate as the default (as previously implemented)?

w-lfchen: i'am pretty sure non-warping rotation is a saner default, the documentation should be updated accordingly

mario-kr:
please do not merge yet, I want to double-check it's correct

mario-kr:
Ok, that was a bit close for comfort, but re-checking what I did was the right decision.

The non-warping version is the one with the previously documented method-order, rotate->translate->scale.

I have rebuild this newest force-pushed commit both with `nix build` and `nix develop -> cargo build --bin eww`, both times yielding the desired result of the contents not being warped.

Sry for the mix-up.

w-lfchen:
thank you for double-checking!

just as a heads-up: there should be no difference in how you build if you execute the commands the way you stated ^^

PR #792, opened 2023-06-05
add `EWW_TIME` magic variable

### LINKED ISSUE #791: [FEATURE] EWW_TIME - builtin clock

It would be nice if EWW provided `EWW_TIME`, similar to `EWW_RAM` or `EWW_CPU`. This would eliminate the only remaining polled shell script from my configuration which probably isn't super important but would be nice.

It would be nice to support multiple time zones, so perhaps rather than a single global `EWW_TIME` variable we would use something like:

```lisp
(deftime NY :zone "America/New_York")
```

This adds another top-level construct which might conflict with existing configs, but I think the `def*` namespace is already pseudo-reserved so hopefully it's fine.

It would be used like this:

```lisp
(defwidget clock
  (box
    (box {NY.year})
    (box {NY.month})
    (box {NY.day})
    (box {NY.hour})
    (box {NY.minute})
    (box {NY.second})))
```

We can provide milliseconds and microseconds for the truly time-obsessed as well (assuming it's cheap) but I don't expect those will get used much.

### Additional context

I can probably put some time in to implementing this, assuming the design outlined above looks OK.

Erenoit:
## Description

Adds new magic variable `EWW_TIME` to get time information. A lot of people use some
type of time information in their widgets in all sorts of ways. I added fields for most
usage cases but I can add more if needed. I think field names are self explanatory but I can
explain one by one if needed.

I used `date` from `coreutils` to get the time information from the system. Therefore, it should
work on most of the `unix` machines. It should also respect `LC_TIME`.

Additionally reduced update delay to 1 second to be able to measure seconds correctly.

This is my first pull request; so, any help is appreciated.

## Usage

It can be used the same as other magic variables. For example:

```
(label :text "${EWW_TIME.day}/${EWW_TIME.month_num}/${EWW_TIME.year}")
```
should output `06/06/2023`.

## Additional Notes

- Not exactly requested feature bu may close #791.
- It is only tested on linux. More test may be needed.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Add Vanilla CSS support (By: Ezequiel Ramis)
 - Add `jq` function, offering jq-style json processing
 - Add `justify` property to the label widget, allowing text justification (By: n3oney)
+- Add `EWW_TIME` magic variable (By: Erenoit)
 
 ## [0.4.0] (04.09.2022)
 
--- a/crates/eww/src/config/inbuilt.rs
+++ b/crates/eww/src/config/inbuilt.rs
@@ -29,7 +29,7 @@ macro_rules! define_builtin_vars {
     }
 }
 
-define_builtin_vars! { Duration::new(2, 0),
+define_builtin_vars! { Duration::new(1, 0),
     // @desc EWW_TEMPS - Heat of the components in Celcius
     // @prop { <name>: temperature }
     "EWW_TEMPS" => || Ok(DynVal::from(get_temperatures())),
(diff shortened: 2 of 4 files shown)


bobtwinkles:
I think we should pull in `chrono` rather than relying on the system's `date` utility, similar to how the other system monitors use the `sysinfo` crate. This approach also doesn't support multiple timezones.

elkowar:
Hmmm, I like the idea of having this!
A few points tho:

I think we should just find some way to have different magic vars have different update intervals (and/or allow them to be push based instead of poll based, for other usecases). Noone needs their disk usage updating every second. 


Secondly:
I'm unsure about having a giant json object for this. What would you think about the following alternative:
Have a single EWW_TIME var which contains a Unix timestamp in seconds, then have simplexpr provide a few functions for date time formatting, I. E.
`formattime(EWW_TIME, "YYYY/MM/DD")` just following whatever format string style chrono or whoever allow. 
That way we could also allow using something like the humantime crate (I think that's what it was) for `relativehumantime(NOTIFICATION_TIMESTAMP)`, outputting "3 minutes ago", and things like that.

This would improve the composability a good bit over a big date json object, as it'd trivially allow doing arithmetic with dates, especially if we added more date related convenience functions to convert seconds, minutes, hours, days etc from and into unix timestamps.


This would also just compose with scripts a lot better, as most things can output Unix timestamps, nothing would output our exact json format.

PR #644, opened 2022-12-10
Clean up example SCSS file

isti115:
Remove some duplicated code and adjust whitespace

PR #816, opened 2023-06-23
Remove duplicate CSS and use .bar class in the example

thled:
I just started using Eww and tried the example. There I noticed small issues and fixed them.

## Description

In the example files I removed two CSS class definitions that where duplicated and used the `.bar` class, which was unused before, on the bar widget.

## Checklist

### DIFF
--- a/examples/eww-bar/eww.yuck
+++ b/examples/eww-bar/eww.yuck
@@ -1,5 +1,5 @@
 (defwidget bar []
-  (centerbox :orientation "h"
+  (centerbox :class "bar" :orientation "h"
     (workspaces)
     (music)
     (sidestuff)))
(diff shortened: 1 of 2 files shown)


examples/eww-bar/eww.yuck
```diff
@@ -1,5 +1,5 @@
 (defwidget bar []
-  (centerbox :orientation "h"
+  (centerbox :class "bar" :orientation "h"
```
w-lfchen, line 2: the `bar` window has the `bar` class, so this isn't needed

w-lfchen (changes requested):
duplicate of #644 
also, the `.metric scale trough highlight` does something and therefore should not be removed

PR #821, opened 2023-06-27
Hacky workarround for log file growing too large

viandoxdev:
Not meant to be merged but an attempt at a workaround for the people who need one. Not tested yet as I can't reproduce the widget deallocation problem anymore.

### DIFF
--- a/crates/eww/src/widgets/def_widget_macro.rs
+++ b/crates/eww/src/widgets/def_widget_macro.rs
@@ -48,8 +48,26 @@ macro_rules! def_widget {
                                 let $gtk_widget = gdk::glib::clone::Downgrade::downgrade(&$gtk_widget);
                                 move |$scope_graph, values| {
                                     // TODO when this fails, shouldn't we technically remove the listener somehow? Need to analyze when exactly this happens.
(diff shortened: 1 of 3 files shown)


elkowar (commented):
Oh boi, this is indeed very hacky.. I still think there should be some proper way of doing log rotation, and / or to just not have this error occur 😂 Is this error the main reason for the spammed logs, to the point where this is actually a "fix"?

viandoxdev:
> Oh boi, this is indeed very hacky.. I still think there should be some proper way of doing log rotation, and / or to just not have this error occur joy Is this error the main reason for the spammed logs, to the point where this is actually a "fix"?

From what I've heard all the very large log files are filled with this "widget got deallocated" error, the best would indeed to not have the error occur at all. I guess we could return some kind of continuation/control flow value from the closure indicating if the listener should be removed, it'd be a little weird API wise (returning `Ok(Continuation::Stop)` on error). Idk if this would be a fix, I'd need a way to test it, but I can't even reproduce the error right now.

elkowar:
This is now fixed in a different way by truncating log files after they grow too large, so I'll close this for now! Thanks, though!

PR #839, opened 2023-07-04
Make transform widgets rotate/scale around their centers

Hyeve-jrs:
## Description

I modified the transform widget to do rotation / scaling around the center of the widget, rather than around 0,0 (usually the top left of the screen).     

### Showcase

This means it now has an effect like this:
image

Rather than this:
image

## Additional Notes

Ideally, I think it'd be useful to base the transform origin off of some property, either new ones or using existing ones like v/h `align`, but I haven't attempted to figure out how to do that. Transforming around the center is fine for my use, and probably significantly more useful than transforming around 0,0 in general.


Hyeve-jrs:
(Just realized that this is effectively a duplicate)

PR #952, opened 2023-10-21
Add jq_raw function to simplexpr

### LINKED ISSUE #745: [FEATURE] Add option to return jq command with raw output

For my use case, I retrieve the active window name from a JSON returned from Hyprland IPC, but since there is no option to get raw output, like with the command line option with `jaq --raw-output ....`, the output is `"somethinginsidequotes"`.

I could remove the quotes by wrapping the jq call inside a replace call that removes the first and last quote characters `"${replace(jq(...), "\(^\"\|\"$\)", "")}"`, but it would be cleaner to have this feature inside jq natively.

Maybe just add alongside the normal `jq` function a `jq_raw` function, as not to break current configurations.

### Additional context

_No response_

RomanHargrave:
## Description
This is a fairly simple change which adds a `jq_raw` function - `jq_raw()` shares `jq()`'s implementation, with the difference being that _top level_ strings are not quoted (nested strings _are_ quoted). This behavior is consistent with `jq`'s documented and observed behavior for --raw-output.

This PR resolves #745.

For instance, we see the following behavior with orthodox `jq`

```
$ echo '{"a": ["b"]}' | jq -r '.a'
[
  "b"
]
$ echo '{"a": ["b"]}' | jq -r '.a[]'
b
```

Similarly, the expressions
```
{jq_raw("{\"a\": [\"b\"]}", ".a")}
{jq_raw("{\"a\": [\"b\"]}", ".a[]")}
```
Produce the following output,

image

## Usage

Use `jq_raw` as you would `jq` when unquoted strings are needed (typically for display)

## Checklist

- [X] I added my changes to CHANGELOG.md, if appropriate.
- [X] The documentation in the `docs/src` directory has been adjusted to reflect my changes.

### DIFF
--- a/crates/simplexpr/src/eval.rs
+++ b/crates/simplexpr/src/eval.rs
@@ -430,8 +430,8 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
             [json] => Ok(DynVal::from(json.as_json_object()?.len() as i32)),
             _ => Err(EvalError::WrongArgCount(name.to_string())),
         },
-        "jq" => match args.as_slice() {
-            [json, code] => run_jaq_function(json.as_json_value()?, code.as_string()?)
+        what @ ("jq" | "jq_raw") => match args.as_slice() {
+            [json, code] => run_jaq_function(json.as_json_value()?, code.as_string()?, what == "jq_raw")
                 .map_err(|e| EvalError::Spanned(code.span(), Box::new(e))),
             _ => Err(EvalError::WrongArgCount(name.to_string())),
         },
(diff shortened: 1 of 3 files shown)


crates/simplexpr/src/eval.rs
```diff
@@ -478,16 +478,23 @@ fn prepare_jaq_filter(code: String) -> Result<Arc<jaq_core::Filter>, EvalError>
...
         .run(jaq_core::Ctx::new([], &inputs), jaq_core::Val::from(json))
-        .map(|x| x.map(Into::<serde_json::Value>::into))
-        .map(|x| x.map(|x| DynVal::from_string(serde_json::to_string(&x).unwrap())))
+        .map(|r| r.map(Into::<serde_json::Value>::into))
+        .map(|r| {
+            r.map(|v| match v {
+                // Per jq docs, "raw-output" behavior simply omits
+                // quotation marks from strings, and outputs what would
+                // otherwise be valid JSON, so this should replicate that.
+                serde_json::Value::String(contents) if raw_string => DynVal::from_string(contents),
```
w-lfchen, line 492:
```suggestion
                serde_json::Value::String(contents) if raw_output => DynVal::from_string(contents),
```

docs/src/expression_language.md
```diff
@@ -51,6 +51,7 @@ Supported currently are the following features:
 	- `arraylength(value)`: Gets the length of the array
 	- `objectlength(value)`: Gets the amount of entries in the object
 	- `jq(value, jq_filter_string)`: run a [jq](https://stedolan.github.io/jq/manual/) style command on a json value. (Uses [jaq](https://crates.io/crates/jaq) internally).
+    - `jq_raw(value, jq_filter_string)`: as above, but top-level strings will not be quoted, as with `jq --raw-output`.
```
w-lfchen, line 54:
```suggestion
	- `jq(value, jq_filter_string)`: run a [jq](https://jqlang.github.io/jq/manual/) style command on a json value. (Uses [jaq](https://crates.io/crates/jaq) internally).
    - `jqraw(value, jq_filter_string)`: acts as if `jq` was invoked with the `--raw-output` flag, removing quotation marks if the resulting value is a JSON string. See the [jq docs](https://jqlang.github.io/jq/manual/) for more information.
```
point to the documentation here to make things easier for us

w-lfchen (changes requested):
looks good, just wrote down some thoughts to consider

elkowar:
Closing this as it was implemented in #1199! Thanks a lot for you effort, and I apologize for the huge delay ._.

PR #1014, opened 2024-01-17
Add custom tooltips

Rayzeq:
## Description

This PR add a widget named `tooltiped`, which is a widget that have eww widgets inside its tooltip instead of only text.

## Usage

For example you can put an image inside a tooltip like this:
```yuck
(tooltiped
    (image :path "/home/....")
    (label :text "example")
)
```

### Showcase

This allow doing thing like this for example:
a

## Checklist

  This folder doesn't exist, I assume you mean `docs/src`, which I don't think I need to update.


elkowar:
Thanks! This is really cool!

PR #1029, opened 2024-02-19
Completions

Rayzeq:
This PR makes it so that the `shell-completions` subcommand is handled before anything else.
This is useful in NixOS when generating shell completions in the build of a derivation. Without this PR, it fails with this error:
```
 2024-02-19T19:37:36.923Z ERROR eww::error_handling_ctx > Failed to initialize eww paths

Caused by:
    Configuration directory /homeless-shelter/.config/eww does not exist
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to eww will be listed here, starting at changes since versio
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


classabbyamp:
can confirm this issue exists in void linux's packaging system and this patch fixes it

PR #1084, opened 2024-04-23
Fix string truncation

Rayzeq:
## Description

Some users have issues because gtk truncate text when it shouldn't, so I've added a `:truncate` property on labels, disabled by default.
As specified in the property documentation, it has no effect is `show-truncated` is `false`, or if `limit-width` has a value.
This behavior allows configurations using the old system (where `limit-width` was mandatory to have truncation) to continue working without the need to add `:truncate true`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ All notable changes to eww will be listed here, starting at changes since versio
 ### Fixes
 - Fix and refactor nix flake (By: w-lfchen)
 
+### Features
+- Add `:truncate` property to labels, disabled by default (except in cases where truncation would be enabled in version `0.5.0` and before) (By: Rayzeq).
+
 ## [0.6.0] (21.04.2024)
 
 ### Fixes
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -918,7 +918,7 @@ fn build_gtk_label(bargs: &mut BuilderArgs) -> Result<gtk::Label> {
 
                 let limit_width = limit_width as usize;
                 let char_count = text.chars().count();
-                if char_count > limit_width && !show_truncated {
+                if char_count > limit_width {
                     if truncate_left {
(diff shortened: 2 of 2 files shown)


AlaisterLeung:
```
(defwidget title_module []
  (box :width 600
    (label :text "qwertyuiopasdfghjkl" :limit-width 10)
  )
)
```
This does not truncate text, but

```
(defwidget title_module []
  (box
    (label :text "qwertyuiopasdfghjkl" :limit-width 10)
  )
)
```
This does truncate the label to 10 characters

Is the width property supposed to work like this? If I want the box to be 600px wide (for background color) and keep the label 10 characters, is this the correct way?

Rayzeq:
According to gtk documentation (`GtkLabel:max-width-chars` is what is used by `:limit-width`):

> The `GtkLabel:width-chars` and `GtkLabel:max-width-chars` properties can be used to control the size allocation of elliptical or wrapped labels. For elliptical labels, if either is specified (and less than the actual text size), it is used as the minimum width, and the actual text size is used as the label's natural width.

From this, it seems that `GtkLabel:max-width-chars` is not a hard limit, so gtk may not truncate the text if there's enough space to display it in full.

I guess you should do it like this if you want a hard limit:
```
(defwidget title_module []
  (box :width 600
    (box
        (label :text "qwertyuiopasdfghjkl" :limit-width 10)
    )
  )
)
```
If you don't care about the ellipsis, you can also put `:show-truncated false` in your label. It won't show the ellipsis, but it will also make `:limit-width` a hard limit.

WantToLearnJapanese:
This has broken the text line wrapping.
There is no conditon where  `gtk_widget.set_ellipsize(pango::EllipsizeMode::None);` and `gtk_widget.set_max_width_chars(limit_width);`  are called at the same.
So, the wrapping text is broken.

Rayzeq:
Yeah `gtk_widget.set_max_width_chars` was never called before, so it hasn't changed anything. I think you also need `gtk_widget.set_wrap(true)` as it's not enabled by default.

PR #1097, opened 2024-05-05
systray: add  property

vnva:
## Description

Added the `:visible-empty` property for the `systray` widget. By default, an empty tray is displayed.

I also noticed a problem that tray elements are not removed from `systray.items`.

## Usage

```yuck
(systray :visible-empty false)
```

### Showcase

https://github.com/elkowar/eww/assets/57404657/d4faf8e2-9bac-4432-978f-c49365a4425a

## Additional Notes

This is my first time writing in rust, so I might have made mistakes, please check carefully to see if I correct any mistakes.

I don’t understand in what format I need to describe the changes in `CHANGELOG.md`, and as far as I understand, the documentation for widgets is generated automatically.

## Checklist

- [ x] I used `cargo fmt` to automatically format all code before committing

### DIFF
--- a/crates/eww/src/widgets/systray.rs
+++ b/crates/eww/src/widgets/systray.rs
@@ -34,12 +34,13 @@ fn run_async_task<F: Future>(f: F) -> F::Output {
 pub struct Props {
     icon_size_tx: tokio::sync::watch::Sender<i32>,
     pub prepend_new: Rc<RefCell<bool>>,
+    pub visible_empty: Rc<RefCell<bool>>,
 }
 
 impl Props {
     pub fn new() -> Self {
         let (icon_size_tx, _) = tokio::sync::watch::channel(24);
-        Self { icon_size_tx, prepend_new: Rc::new(RefCell::new(false)) }
+        Self { icon_size_tx, prepend_new: Rc::new(RefCell::new(false)), visible_empty: Rc::new(RefCell::new(true)) }
     }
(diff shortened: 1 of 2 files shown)


vnva:
Most likely this is not the best solution, I think it’s worth passing a condition on the number of elements in the tray in :visible, you don’t have to merge this PR. But it’s certainly worth fixing removing elements from the tray.

vnva:
@jlo62 i wrote a stupid script to count the number of elements in the tray, you can use it together with `deflisten` and `:visible`

```bash
#!/bin/bash

dbus-monitor --session "interface='org.kde.StatusNotifierWatcher'" |
while read -r signal; do
    if [[ $signal == *"StatusNotifierItemRegistered"* ]]; then
      count=$(($count + 1))
    elif [[ $signal == *"StatusNotifierItemUnregistered"* ]] then
      count=$(($count - 1))
    fi

    echo $count
done

```

vnva:
@elkowar, perhaps this method should be added to the docs or add a magic variable with the number of elements in the systray.

w-lfchen (commented):
i think adding a magic var would be a lot nicer, but i'd have to investigate whether this is difficult from an architectural standpoint.

since there has been little activity on this pr, do you mind me adopting it?
i'd then see where i can take this idea (i'd probably attempt to implement a magic var since that's far more flexible)

Issue #1116, opened 2024-06-10
[BUG] Can't control selected Stack Child using `:selected`


Faervan:
### Checklist before submitting an issue

- [X] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [X] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [X] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

### Description of the bug

I can't get the stack widget to work. It always only shows the first defined child, instead of using the index provided by `selected`.
`eww logs` gives a GtkError:
```
(eww:869899): Gtk-WARNING **: 21:36:31.593: Child name 5 not found in GtkStack
```

### Reproducing the issue

```
; Shows a instead of f
(stack :selected 5 :transition "none" :same-size false
          (label :class "icon" :text "a")
          (label :class "icon" :text "b")
          (label :class "icon" :text "c")
          (label :class "icon" :text "d")
          (label :class "icon" :text "e")
          (label :class "icon" :text "f")
          (label :class "icon" :text "g")
          (label :class "icon" :text "h"))
```

### Expected behaviour

```
A widget that displays one of its children at a time
selected: int index of child which should be shown
```
Should show one child element, depending on the int given to `:selected`

PR #1119, opened 2024-06-18
Fix: the gtk `stack` widget bugfix

ovalkonia:
## Description

The `:selected` property had initially no effect and resulted in a warning stating that a child with the given name wasn't found. This exact issue has already been described here ->  https://github.com/elkowar/eww/issues/1116.

Take a look at the following screenshot showing minimal configuration:

As you can see, there's a warning and instead of the expected "1st", it shows "0th".
Although, if you update the variable, it will work, since all the children are already there:

The problem is here:
https://github.com/elkowar/eww/blob/e95d3af963327f85de86e110fe1514be0c6473bc/crates/eww/src/widgets/widget_definitions.rs#L980-L1011
`gtk_widget.set_visible_child_name(&selected.to_string())` gets called before any children are actually added.

Fixes https://github.com/elkowar/eww/issues/1116

## Usage

N/A

### Showcase

Here is that same minimal configuration from before, but this time using the fixed version:

No warnings and it shows "1st", just as expected!

## Additional Notes

N/A

## Checklist

### DIFF
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -1089,7 +1089,7 @@ const WIDGET_NAME_STACK: &str = "stack";
 fn build_gtk_stack(bargs: &mut BuilderArgs) -> Result<gtk::Stack> {
     let gtk_widget = gtk::Stack::new();
 
-    if let Ordering::Less = bargs.widget_use.children.len().cmp(&1) {
+    if bargs.widget_use.children.len() < 1 {
         return Err(DiagError(gen_diagnostic!("stack must contain at least one element", bargs.widget_use.span)).into());
     }
 
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -1088,6 +1088,27 @@ const WIDGET_NAME_STACK: &str = "stack";
 /// @desc A widget that displays one of its children at a time
 fn build_gtk_stack(bargs: &mut BuilderArgs) -> Result<gtk::Stack> {
     let gtk_widget = gtk::Stack::new();
+
+    if let Ordering::Less = bargs.widget_use.children.len().cmp(&1) {
+        return Err(DiagError(gen_diagnostic!("stack must contain at least one element", bargs.widget_use.span)).into());
+    }
+
+    let children = bargs.widget_use.children.iter().map(|child| {
+        build_gtk_widget(
+            bargs.scope_graph,
+            bargs.widget_defs.clone(),
+            bargs.calling_scope,
+            child.clone(),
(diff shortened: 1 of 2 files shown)


fabolous005:
Hi,
I'm also currently looking into the stack widget. So far with no luck.
I thought i could index the childs of the stack like this:
```yuck
(defvar stack-index 0)

(stack
	:selected stack-index
        (child-widget1)
        (child-widget2)
)
```
This would be extremely useful because you could update the variable from `on-hover` events from `eventbox`!
Do you know how to archive this or if it is even possible (yet)?

ovalkonia:
@fabolous005 Yeah, absolutely you can  can do that! `Eww` is quite flexible, the possibilities are endless. Try something like this:

```lisp
(defvar stack-index 0)

(eventbox
        :onhover "${EWW_CMD} update stack-index=${(stack-index + 1) % 3}"
        (stack
            :selected stack-index
            (label :text "first")
            (label :text "second")
            (label :text "third")
        )
 )
```

It's just a `stack` wrapped inside of an `eventbox`. If you hover over, it will switch to the next child.

You can update variables using `eww update variable=value`, and `EWW_CMD` is just a shortcut for the current running configuration (see https://elkowar.github.io/eww/magic-vars.html#eww_cmd).

For `eventbox` docs, see https://elkowar.github.io/eww/widgets.html#eventbox. Hope it helped!

fabolous005:
Yeah, i tried that, but whatever i try i get the error `Child name '0' not found`, even with your exact example.
evv version: `eww 0.6.0 4d55e9ad63d1fae887726dffcd25a32def23d34f`

ovalkonia:
> Yeah, i tried that, but whatever i try i get the error `Child name '0' not found`, even with your exact example. evv version: `eww 0.6.0 4d55e9ad63d1fae887726dffcd25a32def23d34f`

Well, it is what it is for now, this pr solves this exact issue. Right now it essentially defaults to the `0th` child. If you update the variable, it will work since all the children will be there. Just ignore the error message and try hovering over it, it should work.

crates/eww/src/widgets/widget_definitions.rs
```diff
@@ -1088,6 +1088,27 @@ const WIDGET_NAME_STACK: &str = "stack";
 /// @desc A widget that displays one of its children at a time
 fn build_gtk_stack(bargs: &mut BuilderArgs) -> Result<gtk::Stack> {
     let gtk_widget = gtk::Stack::new();
+
+    if let Ordering::Less = bargs.widget_use.children.len().cmp(&1) {
```
w-lfchen, line 1092:
```suggestion
    if bargs.widget_use.children.len() < 1 {
```

w-lfchen (changes requested):
you can probably just apply this from the web interface
apart from this minor nitpick lgtm

PR #1120, opened 2024-06-18
Remove window flicker and extend duration with open command

### LINKED ISSUE #892: [BUG] Opening a window that is already open closes and reopens it, causing flickering

### Description of the bug

Running `eww open` on a window that is already open will cause the window to close and then reopen causing flickering. Seems to had been fixed a while ago in https://github.com/elkowar/eww/pull/263 but it still seems to be an issue.

### Reproducing the issue

Running `eww open` on an already open window will cause it, however I noticed it when using the new `--duration` flag.

### Expected behaviour

Opening a window that is already open should just keep the window open without first closing it.
In the case of the `--duration` flag, it should just reset the duration the window should be open for.

### Additional context

Noticed it while using eww to create a custom volume/brightness OSD. Running the `eww open osd --duration 2000` command I noticed the flickering, and after checking eww logs it confirmed my suspicion. If there is a better way to do this, excuse my ignorance with this bug report.

BeneSim:
## Description

This solves a problem that had been already solved in 0d1b5e0f9e532ade4b7c9fd77b46b6a29b1d8529 but was reintroduced in a9a35c1804d72ef92e04ee71555bd9e5a08fa17e. Basically before this, the window has been closed and reopened if it existed, which results in flickering due to the animations.

In addition to removing the window flicker the timer (if started with --duration) will also be renewed. This fixes #892.

## Additional Notes

- This is a breaking change, people who relied on using `open` to apply new settings to a window will be surprised. However, in 0d1b5e0f9e532ade4b7c9fd77b46b6a29b1d8529 it was mentioned that this could and should be solved by `reload`.
- With this implementation the window close timer won't be aborted if no `--duration` has been supplied, e.g.: I open a window with `--duration 4s` after 3s i open the window again but without `--duration` -> The window will still close after 1s. This might be counter-intuitive and could be fixed quite easily, I just wasn't sure what behavior to implement.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 ## Unreleased
 
 ### Fixes
+- Fix window flicker when opening an already existing window (By: benesim)
 - Fix and refactor nix flake (By: w-lfchen)
 - Fix remove items from systray (By: vnva)
 
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -362,14 +362,46 @@ impl<B: DisplayBackend> App<B> {
         Ok(())
     }
 
+    fn setup_close_timer(&mut self, instance_id: &str, duration: Duration) {
+        let app_evt_sender = self.app_evt_send.clone();
+
+        let (abort_send, abort_recv) = futures::channel::oneshot::channel();
+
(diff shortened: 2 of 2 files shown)


elkowar:
I think if the window is re-opened with no duration provided, the timer should be canceled, that seems like the most intuitive behavior to me.

Issue #1040, opened 2024-02-27
[BUG] rgba is limited from :style in yuck


ArtemSmaznov:
### Checklist before submitting an issue

- [X] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [X] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [X] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

### Description of the bug

Is there a reason why rgba() behaves differently between cscc and yuck?

This works fine:
```scss
.music-player {
    background: rgba(#FF0000, 0.4);
}
```

This for some reason doesn't:
```yuck
(box :class "music-player"
     :style "background: rgba(#FF0000, 0.4);"
```

It works fine without the opacity though:
```yuck
(box :class "music-player"
     :style "background: #FF0000;"
```

To clarify what I am trying to do.
I have a var `music-player-album-color` in yuck and I want to make the music player widget have a dynamic color AND be semi-transparent. I can get to having either one of those but not both.

I have even tried setting a new scss variable from yuck but it seems anything set there operates on a different scope than what I write in scss so couldn't get this working as a workaround.

Is there a bug in how rgba() in handled from yuck or is this a limitation that is easy to overcome?

### Reproducing the issue

Add the following to any widget
```
     :style "background: rgba(#FF0000, 0.4);"
```

### Expected behaviour

the widget should get colored accordingly (40% transparent red)

ArtemSmaznov:
Thanks @ovalkonia for the explanation, it is very informative.
I didn't want to use `:css` property as that would require me to specify the class again and I am not a fan of needless duplication.
For the time being I just changed my script that sets eww variable to output color in the `255, 0, 0` format instead of `#FF0000` one and it is working for me with the `:style` propery. I would like to still use the hex format though with `:style` if possible.
Will give `:css` a try though when I get the chance to see if I'd like that.

ovalkonia:
I don't think you are actually required to specify the class again with the `:css` property. Take a look at the code that handles both `:css` and `:style` properties:
![code](https://github.com/elkowar/eww/assets/60359793/9034b528-ab44-406a-a54a-80276fc84beb)
[...]

PR #1121, opened 2024-06-18
Add scss support for style property

ovalkonia:
## Description

Instead of applying the raw data at once in the `:style` widget property, it now feeds it to the `grass::from_string` function before applying, just as it's done in the `:css` property. It makes the usage of SCSS features possible, and since SCSS is compatible with CSS, there's no breaking changes.

Closes https://github.com/elkowar/eww/issues/1040

## Usage

N/A

### Showcase

Now this minimal configuration works just fine, which would not have gone so smoothly before due to the `rgba(#FF0000, 0.4)` format being a SCSS feature only!

## Additional Notes

N/A

## Checklist


w-lfchen (approved):

elkowar:
While there are technically some things in GTK-CSS that _are_ incompatible with scss, I think they are uncommon enough that it's reasonable to assume it won't come up. I'll merge this, although in the unlikely case that people _do_ rely on some niche GTK-css specific syntax somewhere in these style attributes, we'll need to revert back.

Issue #1086, opened 2024-04-25
[BUG] Eww uses monitor model instead of a serial


coderPaddyS:
### Checklist before submitting an issue

- [X] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [X] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [X] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

### Description of the bug

Spawning a widget on each monitor is only possible, if they aren't the same model. If they are, eww will spawn the widget on the same monitor. Using the monitor serial primarily and the model as fallback (or any property specified by the user) or the monitor description would resolve this issue.
Usage of an index as id could be a solution, but sometimes the monitor ordering is not the same as reported by the DE (in my case Hyprland) as seen below. Eww uses the same ordering in which my DE lists the monitors, not by the ID.

### Reproducing the issue

```
$>eww open bar --id 1 --arg screen=609NTFAF4483
failed to open window `1`

Caused by:
    Failed to get monitor 609NTFAF4483
    The available monitors are:
        [0] 0x1551
        [1] 24EB23
        [2] 24EB23
```
The monitor ordering:
```
$> hyprctl monitors
Monitor eDP-1 (ID 0):
        3840x2400@59.99400 at 960x1200
        description: Sharp Corporation 0x1551
        make: Sharp Corporation
        model: 0x1551
        serial:

Monitor DP-8 (ID 2):
        1920x1200@59.95000 at 1920x0
        description: LG Electronics 24EB23 305NDGL87831
        make: LG Electronics
        model: 24EB23
        serial: 305NDGL87831


Monitor DP-7 (ID 1):
        1920x1200@59.95000 at 0x0
        description: LG Electronics 24EB23 609NTFAF4483
        make: LG Electronics
        model: 24EB23
        serial: 609NTFAF4483

### Expected behaviour

Eww spawns a window on each monitor correctly by specifying the serial, model or description of the monitor.

### Additional context

![2024-04-25T15:14:54,802040489+02:00](https://github.com/elkowar/eww/assets/48881352/92559db8-2d19-4663-a6e4-be5ff4b18917)

krims0n32:
[...]

Issue #875, opened 2023-08-15
[FEATURE] Allow output names for wayland monitors


hiszd:
### Description of the requested feature

Essentially this would allow someone to assign a window to a display by inputting the name of the monitor(e.g. "eDP-1") instead of the monitor number.

### Proposed configuration syntax

The syntax for this would look like this
```
:monitor "eDP-1"
```

### Additional context

This is basically the same as PR #522 but for Wayland.

viandoxdev:
As discussed previously on that very PR, this is really hard to do properly since GTK doesn't actually expose that info. I'll have to look if things have changed on the gtk-rs side and come back with an update.

luposmi:
@viandoxdev are you still working on this?
I made a quick lookup and 
```rust
    let monitor: gdk::Monitor = monitors.item(0).unwrap().downcast().unwrap();
    println!("Monitor Name: {:?}", monitor.description());
```
    prints "Monitor Name: Some("AOC 2369M AJDG89A003953 (HDMI-A-1)")"
    which you could get the name from
otherwise, i would have time next week to make a pr

viandoxdev:
> @viandoxdev are you still working on this?
> I made a quick lookup and 
> ```rust
>     let monitor: gdk::Monitor = monitors.item(0).unwrap().downcast().unwrap();
>     println!("Monitor Name: {:?}", monitor.description());
> ```
>     prints "Monitor Name: Some("AOC 2369M AJDG89A003953 (HDMI-A-1)")"
>     which you could get the name from
> otherwise, i would have time next week to make a pr

I'm not particularly interested in implementing this so don't hesitate to make a PR. I'm not convinced you can reliably extract the name from that description (if that's what you intend to do), but it would be great if you could manage, good luck

luposmi:
Still working on it, but i am quite busy. Might take two weeks.

luposmi:
Did a little more digging and GDK gets its information from https://wayland.app/protocols/xdg-output-unstable-v1#zxdg_output_v1:event:description whose output differs by compositor. In gtk4, it would be possible to get this information from wl_output, so, 

tldr: not really possible at the moment without very janky solutions.

w-lfchen:
should be fixed by 510b824e7545a7c98e050522a7bc93d884c53432 which is indeed very janky

PR #1129, opened 2024-07-01
Use gdk_screen_get_monitor_plug_name to provide more consistent monitor names on wayland

dragonnn:
## Description

This fixes issue https://github.com/elkowar/eww/issues/1086 and https://github.com/elkowar/eww/issues/875 (at least partly).
It is based on the linked ags issue in 1086. Yes it is a bad workaround with does require unsafe, pulling in gdk-sys and using a deprecated function from it but really without it eww is basically not usable on wayland with any kind of dynamic monitor setup (common when using laptops).

## Usage

Allows using monitor plug names under wayland like that:
```
(defwindow bar-DP-2
  :monitor "DP-2"
  :geometry (geometry :width "100%"
    :anchor "bottom center"
  )
  :exclusive true
  :namespace "eww"
  (bar)
)
```

## Additional Notes

Even that it uses unsafe, deprecated functions and gdk-sys I really think this is worth adding since as far I have researched this topic they is no other way of improving wayland monitor support other then rewriting to gtk4 with I suspect doesn't happen soon. 
Tested it on niri windows manager and works fine.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Fix the gtk `stack` widget (By: ovalkonia)
 - Fix values in the `EWW_NET` variable (By: mario-kr)
 - Fix the gtk `expander` widget (By: ovalkonia)
+- Fix wayland monitor names support (By: dragonnn)
 
 ### Features
 - Update rust toolchain to 1.80.1 (By: w-lfchen)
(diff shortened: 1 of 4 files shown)


w-lfchen:
i've spent the past 4 hours looking into a better implementation for this but i sadly have to give up due to my lack of knowledge on this topic.
i think there is some way using the `wayland_client` and `wayland_protocols` crates, we need to get the compositor to list its outputs somehow.
i think [this protocol](https://wayland.app/protocols/xdg-output-unstable-v1) is the one that we need, but i'm not 100% sure about this.
the `name` property is the one we're looking for.

Issue #1085, opened 2024-04-23
[BUG] Expander widget does not expand if name is specified


Life-Not-Found:
### Checklist before submitting an issue

- [X] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [X] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [X] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

### Description of the bug

If a name is specified in the expander widget the expander will not expand

https://github.com/elkowar/eww/assets/113875139/d085bbcd-2901-433a-8288-021d37a49f4f

### Reproducing the issue

```
(
            expander
            :name "Name of expander"
            (
                label
                :text "expanding"
            )
        )
```

### Expected behavior

The expander expands and shows child widget

### Additional context

Was able to replicate in a VM

https://github.com/elkowar/eww/assets/113875139/b1b88b45-f229-48f6-9555-bf31c23f543e

_No response_

PR #1132, opened 2024-07-04
Fix: the gtk `expander` widget bugfix

ovalkonia:
## Description

As it was mentioned in this issue -> https://github.com/elkowar/eww/issues/1085, if the `name` property was set for the `expander` widget, then child elements didn't show up. Essentially it was a choice between a labeled but child-free `expander` or labeless but with children `expander`.

This behavior came from the `build_builtin_gtk_widget()` function. To be more precise, this function adds child elements only if there aren't any children that have been already added with the `widget_use_to_gtk_widget()` function. The problem with that in this case is that `expander`'s label is actually considered to be a child element itself:

So now the `build_gtk_expander()` function handles the child element explicitly (see diff).

Fixes https://github.com/elkowar/eww/issues/1085

## Usage

N/A

### Showcase

Here's the minimal configuration used below:
```
(defwindow example
    :monitor "0"
    :geometry (geometry
        :x "0"
        :y "0"
        :width "100%"
        :height "35px"
        :anchor "top center"
    )
    :exclusive "true"
    (expander
        :name "foo"
        :halign "start"
        (label :text "expanded child")
    )
)
```

The result:

Works as expected

## Additional Notes

Now the `expander` widget is also required to have exactly one child. If no children were provided, there would be an error. I just thought it would be nice, but it can be removed without any problems if necessary.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Fix nix flake
 - Fix `jq` (By: w-lfchen)
 - Labels now use gtk's truncation system (By: Rayzeq).
+- Fix the gtk `expander` widget (By: ovalkonia)
 
 ### Features
 - Add `systray` widget (By: ralismark)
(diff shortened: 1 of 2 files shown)


w-lfchen (approved):

elkowar:
Thanks for your work! Merged!

ovalkonia:
@elkowar, whoops, I think I accidentally put the changelog entry under the wrong version. Should be under `unreleased`, my bad. Must have been sleepy when I did this. Can you fix that, please? Another pr would probably be too much

w-lfchen:
i've added this to #1155, that pr contains some general fixes by now

ovalkonia:
Thank you!

PR #1137, opened 2024-07-15
Update GTK and related dependencies to 0.18

geluk:
## Description

This PR updates GTK, GDK, and all related dependencies to 0.18.

Additionally, since `gtk` re-exports `gdk`, `gdk-pixbuf`, `glib`, `pango`, and `cairo`, it is possible (and I'd argue preferable) to depend directly on the re-exported versions. This way, we don't have to worry about keeping the versions aligned anymore - Cargo will take care of that. Notably, the re-exported version of `gdk-pixbuf` was already used everywhere, so it could just be removed without any import changes.

## Motivation

I'd love to see `eww` updated to GTK 4, but I understand that this is complex and time consuming, and while I wouldn't mind assisting, I don't have enough experience with GTK to be of much use.

That said, what I can at least do is get the 0.17->0.18 breaking changes out of the way, and simplify the dependencies a bit to make the eventual migration ever so slightly easier.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1000,7 +1000,6 @@ dependencies = [
  "extend",
  "futures",
  "gdkx11",
- "glib-macros",
  "grass",
  "gtk",
  "gtk-layer-shell",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,7 +140,7 @@ version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c48ccdbf6ca6b121e0f586cbc0e73ae440e56c67c30fa0873b4e110d9c26d2b"
 dependencies = [
- "event-listener",
+ "event-listener 2.5.3",
  "futures-core",
 ]
 
@@ -151,7 +151,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "81953c529336010edd6d8e358f886d9581267795c61b19475b71314bffa46d35"
 dependencies = [
  "concurrent-queue",
- "event-listener",
+ "event-listener 2.5.3",
  "futures-core",
 ]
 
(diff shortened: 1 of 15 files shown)


yuki0iq:
glib-macros dependency is redundant, as its contents are reexported by glib crate

geluk:
Well spotted, fixed.

w-lfchen:
cargo wants to add 4 dependencies for some reason when using `cargo update`:
```
      Adding gdk-pixbuf-sys v0.20.0
      Adding gio-sys v0.20.0
      Adding glib-sys v0.20.0
      Adding gobject-sys v0.20.0
```
adding them (specifically, `gio-sys`, which seems to be required by `gdk-pixbuf-sys`, adding the other two on their own is fine) breaks `dbusmenu-glib`:
```
error[E0277]: the trait bound `menuitem::Menuitem: glib::translate::FromGlibPtrArrayContainerAsVec<_, *mut glib_sys::GList>` is not satisfied
```
i am at a loss here, do you have any idea what could be causing this?
or should i file an issue against the `dbusmenu-glib` repo?

w-lfchen:
#1144 has been merged, this pr can now be closed

w-lfchen (changes requested):
https://github.com/elkowar/eww/pull/1144 has been merged, this pr can now be closed

elkowar:
Thanks a lot for your work, @geluk! closing as this has been included in #1144

elkowar:
Regarding GTK4:
I even already had a working build of GTK4 eww a couple of years ago. Sadly, there where some showstoppers that I don't think have been fully resolved yet, especially regarding wayland compatibility (GTK4 makes rendering on layershell surfaces a lot more complex, and back then it was fully impossible), as well as some other larger architectural changes that affected eww quite negatively. I don't have any concrete plans on when to retry a GTK4 migration currently, as I think most of the features GTK4 adds don't really affect eww directly.

PR #1146, opened 2024-08-05
feat: Add fill attribute to image widget for custom/dynamic svg color

hypernova7:
## Description
Added fill attribute to image widget for custom/dynamic svg color.

> [!NOTE]
> This option is only for svg images

## Usage

```lisp
(image
  :fill "green"
  :path "./my-svg-icon.svg"
)
```

```lisp
(image
  :fill "#0389ff"
  :path "./my-svg-icon.svg"
)
```

### Showcase
<table>
<tr>

</tr>
<tr>

</tr>
</table>

## Additional Notes

Anything else you want to add, such as remaining questions or explanations.

## Checklist

### DIFF
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -541,7 +541,7 @@ fn svg_to_pixbuf(
         reg.replace(&svg_data, &format!("fill=\"{}\"", fill))
     } else {
         let reg = regex::Regex::new(r"<svg")?;
-        reg.replace(&svg_data, &format!(" fill=\"{}\"", fill))
+        reg.replace(&svg_data, &format!("<svg fill=\"{}\"", fill))
     };
 
     let pixbuf_svg = gtk::gdk_pixbuf::PixbufLoader::with_type("svg")?;
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -546,15 +546,11 @@ fn svg_to_pixbuf(
 
     let pixbuf_svg = gtk::gdk_pixbuf::PixbufLoader::with_type("svg")?;
 
-    if image_width >= 0 && image_height == -1 {
-        pixbuf_svg.set_size(image_width, image_width);
-    }
-    if image_width == -1 && image_height >= 0 {
-        pixbuf_svg.set_size(image_height, image_height);
-    }
-    if image_width >= 0 && image_height >= 0 {
-        pixbuf_svg.set_size(image_width, image_height);
-    }
+    match (image_width, image_height) {
+        (w, h) if w > 0 || h > 0 => pixbuf_svg.set_size(w, h),
+        // Add default size to prevent widget overflowing, if image is too big
+        _ => pixbuf_svg.set_size(24, 24),
+    };
 
     let svg_buf: Vec<u8> = svg_data.as_bytes().to_vec();
     pixbuf_svg.write(&svg_buf)?;
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -574,7 +574,7 @@ fn build_gtk_image(bargs: &mut BuilderArgs) -> Result<gtk::Image> {
         // @prop image-height - height of the image
         prop(path: as_string, image_width: as_i32 = -1, image_height: as_i32 = -1, fill: as_string = "currentColor") {
             if fill != "currentColor" && !path.ends_with(".svg") {
-                log::warn!("The fill sttribute is only for SVG images");
+                log::warn!("The fill attribute is only for SVG images");
             }
 
             if path.ends_with(".gif") {
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -527,38 +527,6 @@ fn parse_icon_size(o: &str) -> Result<gtk::IconSize> {
     }
 }
 
-fn svg_to_pixbuf(
-    path: std::path::PathBuf,
-    image_width: i32,
-    image_height: i32,
-    fill: &str,
-) -> Result<Option<gtk::gdk_pixbuf::Pixbuf>> {
-    let svg_data: String = std::fs::read_to_string(path)?;
-
-    // The fastest way to add/change fill color
-    let svg_data = if svg_data.contains("fill=") {
-        let reg = regex::Regex::new(r#"fill="[^"]*""#)?;
-        reg.replace(&svg_data, &format!("fill=\"{}\"", fill))
-    } else {
-        let reg = regex::Regex::new(r"<svg")?;
-        reg.replace(&svg_data, &format!("<svg fill=\"{}\"", fill))
-    };
-
-    let pixbuf_svg = gtk::gdk_pixbuf::PixbufLoader::with_type("svg")?;
-
-    match (image_width, image_height) {
-        (w, h) if w > 0 || h > 0 => pixbuf_svg.set_size(w, h),
-        // Add default size to prevent widget overflowing, if image is too big
-        _ => pixbuf_svg.set_size(24, 24),
-    };
-
-    let svg_buf: Vec<u8> = svg_data.as_bytes().to_vec();
-    pixbuf_svg.write(&svg_buf)?;
-    pixbuf_svg.close()?;
-
-    Ok(pixbuf_svg.pixbuf())
-}
-
 const WIDGET_NAME_IMAGE: &str = "image";
 /// @widget image
 /// @desc A widget displaying an image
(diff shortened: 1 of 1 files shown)


crates/eww/src/widgets/widget_definitions.rs
```diff
@@ -536,10 +572,17 @@ fn build_gtk_image(bargs: &mut BuilderArgs) -> Result<gtk::Image> {
         // @prop path - path to the image file
         // @prop image-width - width of the image
         // @prop image-height - height of the image
-        prop(path: as_string, image_width: as_i32 = -1, image_height: as_i32 = -1) {
+        prop(path: as_string, image_width: as_i32 = -1, image_height: as_i32 = -1, fill: as_string = "currentColor") {
+            if fill != "currentColor" && !path.ends_with(".svg") {
+                log::warn!("The fill sttribute is only for SVG images");
```
w-lfchen, line 577:
```suggestion
                log::warn!("The fill attribute is only for SVG images");
```

crates/eww/src/widgets/widget_definitions.rs
```diff
@@ -527,6 +527,42 @@ fn parse_icon_size(o: &str) -> Result<gtk::IconSize> {
     }
 }
 
+fn svg_to_pixbuf(
+    path: std::path::PathBuf,
+    image_width: i32,
+    image_height: i32,
+    fill: &str,
+) -> Result<Option<gtk::gdk_pixbuf::Pixbuf>> {
```
w-lfchen, line 535: this api seems a bit weird
why return a `Result<Option<_>>` here? if the operation fails, just return an error

crates/eww/src/widgets/widget_definitions.rs
```diff
@@ -527,6 +527,42 @@ fn parse_icon_size(o: &str) -> Result<gtk::IconSize> {
...
+
+    if image_width >= 0 && image_height == -1 {
+        pixbuf_svg.set_size(image_width, image_width);
+    }
+    if image_width == -1 && image_height >= 0 {
+        pixbuf_svg.set_size(image_height, image_height);
+    }
+    if image_width >= 0 && image_height >= 0 {
+        pixbuf_svg.set_size(image_width, image_height);
+    }
```
w-lfchen, line 557: consider using a match statement to avoid logic errors

w-lfchen (changes requested):
breaks image functionality for [this svg](https://logowiki.net/uploads/logo/g/github-icon.svg) (the first one i tried) without even using the new attribute
the image disappears when using the new function, same behavior as a wrong path being supplied

crates/eww/src/widgets/widget_definitions.rs
hypernova7: I did it that way to avoid using unwrap :laughing:, and also because the pixbuf() method of PixbufLoader can return None if the image is empty.

crates/eww/src/widgets/widget_definitions.rs
```diff
@@ -527,6 +527,38 @@ fn parse_icon_size(o: &str) -> Result<gtk::IconSize> {
...
+        reg.replace(&svg_data, &format!("<svg fill=\"{}\"", fill))
+    };
+
+    let pixbuf_svg = gtk::gdk_pixbuf::PixbufLoader::with_type("svg")?;
+
+    match (image_width, image_height) {
+        (w, h) if w > 0 || h > 0 => pixbuf_svg.set_size(w, h),
+        // Add default size to prevent widget overflowing, if image is too big
+        _ => pixbuf_svg.set_size(24, 24),
+    };
```
w-lfchen, line 553:
```suggestion
    pixbuf_svg.set_size(image_width, image_height);
```
assuming defaults here might break configs relying on image sizes, i suggest removing this.

crates/eww/src/widgets/widget_definitions.rs
w-lfchen: i think it'd be better to move that away from this function then and have it only handle the svg manipulation then
maybe even consider inlining this function entirely

w-lfchen (changes requested):
i think the fill logic should only be invoked if the attribute is set, otherwise defaulting to the already existing code seems like a better option for performance reasons
in order to achieve that, it might help to mark the attribute as optional instead of relying on a default value
also, don't forget to document everything ^^

crates/eww/src/widgets/widget_definitions.rs
w-lfchen: i think it'd be better to move that away from this function then and have it only handle the svg manipulation then
maybe even consider inlining this function entirely

crates/eww/src/widgets/widget_definitions.rs
```diff
@@ -527,6 +527,38 @@ fn parse_icon_size(o: &str) -> Result<gtk::IconSize> {
...
+
+    let svg_buf: Vec<u8> = svg_data.as_bytes().to_vec();
+    pixbuf_svg.write(&svg_buf)?;
```
w-lfchen, line 556:
```suggestion
    pixbuf_svg.write(svg_data.as_bytes())?;
```

crates/eww/src/widgets/widget_definitions.rs
hypernova7: The default values for image_width and image_height are -1, if left as default without adding the declarations, the image will cause the widget to grow and possibly fill the entire screen, resulting in the user no longer being able to manipulate the widget as SVGs tend to inherit size from the parent element if they do not have the width and height attributes.

crates/eww/src/widgets/widget_definitions.rs
hypernova7: I didn't understand, you mean to move the svg_to_pixbuf functionality inside the image widget block where the SVG is handled. Sorry for my bad English.

crates/eww/src/widgets/widget_definitions.rs
hypernova7: The default values ​​for image_width and image_height are -1, if left as default without adding the declarations, the image will cause the widget to grow and possibly fill the entire screen, resulting in the user no longer being able to manipulate the widget as SVGs tend to inherit or from viewBox sizes from the parent element if they do not have the width and height attributes.

crates/eww/src/widgets/widget_definitions.rs
w-lfchen: yes

crates/eww/src/widgets/widget_definitions.rs
w-lfchen: yes, but it'd still be a breaking change and why only implement this logic for svgs and not all images then?
also, why choose exactly 24*24 pixels?

hypernova7: That's true, although currently the image-width and image-height attributes are not used for gif type images, which is wrong, but that would be in a separate PR, don't you think?

hypernova7: That's true, although currently for gif type images the attributes image-width and image-height are not used, which is wrong, but that would be in a separate PR, don't you think?

And I used 24x24 to add a default size in case the attributes image-width/image-height are not defined, since otherwise some svg sizes like the github logo you attached are too big and the widget grows and overflows the screen

I currently added some logic to determine the size of the image in case any of the two attributes are defined, and thus display the image with its correct dimensions

w-lfchen: the attributes not being used with gifs is indeed a separate concern.
i'd rather add a notice to the docs to be careful about using large images instead of breaking the current behavior that people might rely on.
i don't think you understood my other point about the default dimensions
why 24*24 and not, for example, 300*300 as a default?
is there a reason for this?

hypernova7: Yes, and it is to avoid the widget overflowing, the smaller the image dimensions the better, so we avoid future problems. Although you are right, maybe I will change it to 128x128 (which is the maximum standard recommended by most gnome icon packs, as far as I know), but if the image dimensions are less than 128x128 I will use the real dimensions, so there will be no problems.

I appreciate your patience

hypernova7: > it might help to mark the attribute as optional
The image (regardless of the image type gif, webp, etc) is not displayed if I leave fill as optional, that's why I put an empty value as the default value. I'm even thinking that the best thing would be to create an exclusive widget for svg

hypernova7:
By the way, I forgot to mention that I previously made the mistake of defining image_height and instead I set image_width twice, which made me believe that set_size ignored the height value but that is not the case, and set_size is only set if both fields are greater than 0, so I added code to preserve the aspect ratio in case one of the two attributes is used.

Sorry for the delay of this message, I've been busy

hypernova7:
That's the result with last code update


2024-08-06_05-25-36_area_maim

Example
```lisp
(box
  :orientation "v"
  (image :fill "red" :image-height 36 :path "./src/rect.svg")
  (image :image-height 36 :path "./src/rect.svg")
)
```

`rect.svg`
```xml
<svg fill="blue" width="130" height="300" viewBox="0 0 130 300" xmlns="http://www.w3.org/2000/svg">
  <rect width="130" height="300" x="0" y="0" rx="20" ry="20" />
</svg>
```

crates/eww/src/widgets/widget_definitions.rs
w-lfchen: i managed to set it as optional without issues.
a separate might be an idea, but it should be for all kinds of vector graphics and i personally don't really see a reason for it

w-lfchen:
the following code produces two differently sized icons.
```
      (box
        :space-evenly false
        (image
          :fill "#cba6f7"
          :image-height 20
          :image-width 200
          :path "../github-icon.svg"
        )
        (image
          ;:fill "#0000ff"
          :image-height 20
          :image-width 200
          :path "../github-icon.svg"
        ))
```
also, i've looked into the gtk docs for this now, using the pixbufloader is not necessary here.
i'll provide a code suggestion for what i'd consider a simpler solution

crates/eww/src/widgets/widget_definitions.rs
```diff
@@ -536,10 +536,55 @@ fn build_gtk_image(bargs: &mut BuilderArgs) -> Result<gtk::Image> {
...
+                        (w, h) if w >= 0 && h >= 0 => pixbuf_svg.set_size(w, h),
+                        _ => {}
+                    };
+                });
+                pixbuf_svg.write(svg_data.as_bytes())?;
+                pixbuf_svg.close()?;
+
+                gtk_widget.set_from_pixbuf(pixbuf_svg.pixbuf().as_ref());
             } else {
                 let pixbuf = gtk::gdk_pixbuf::Pixbuf::from_file_at_size(std::path::PathBuf::from(path), image_width, image_height)?;
```
w-lfchen, line 589:
```suggestion
            } else {
                let pixbuf;
                // populate the pixel buffer
                if path.ends_with(".svg") && !fill.is_empty() {
                    let svg_data = std::fs::read_to_string(std::path::PathBuf::from(path.clone()))?;
                    // The fastest way to add/change fill color
                    let svg_data = if svg_data.contains("fill=") {
                        let reg = regex::Regex::new(r#"fill="[^"]*""#)?;
                        reg.replace(&svg_data, &format!("fill=\"{}\"", fill))
                    } else {
                        let reg = regex::Regex::new(r"<svg")?;
                        reg.replace(&svg_data, &format!("<svg fill=\"{}\"", fill))
                    };
                    let stream = gtk::gio::MemoryInputStream::from_bytes(&gtk::glib::Bytes::from(svg_data.as_bytes()));
                    pixbuf = gtk::gdk_pixbuf::Pixbuf::from_stream_at_scale(&stream, image_width, image_height, true, None::<&gtk::gio::Cancellable>)?;
                    stream.close(None::<&gtk::gio::Cancellable>)?;
                } else {
                    pixbuf = gtk::gdk_pixbuf::Pixbuf::from_file_at_size(std::path::PathBuf::from(path), image_width, image_height)?;
                }

```

w-lfchen (commented):
this code results in the same scaling behavior when using and not using the fill attribute.
it also allows for easier maintenance, should, for example, a `preserve-aspect-ratio` property be added.

crates/eww/src/widgets/widget_definitions.rs
hypernova7: Nice implementation, 100x better than mine and simpler

hypernova7: I think it would be best to close this PR and for you to submit a new one, since you did much better and the truth is that I just wanted to give an idea since I am very new to this.

crates/eww/src/widgets/widget_definitions.rs
hypernova7: How mark option as optional?

I did it this way but it doesn't work, i.e. the image without the fill attribute is not displayed.


```rs
prop(path: as_string, image_width: as_i32 = -1, image_height: as_i32 = -1, fill: as_string) {
```

crates/eww/src/widgets/widget_definitions.rs
w-lfchen: i can open a new pr if you want, but give yourself some credit too, i don't know anything about the regex stuff you're doing there and the idea to add this never crossed my mind ^^'
i'll also look into the aspect ratio attribute functionality since both concern the same widget so give me a moment to evaluate that

crates/eww/src/widgets/widget_definitions.rs
w-lfchen:
```rust
prop(path: as_string, image_width: as_i32 = -1, image_height: as_i32 = -1, fill: as_string?)
```

crates/eww/src/widgets/widget_definitions.rs
hypernova7: The aspect ratio was just to simulate what Pixbuf::from_file_at_size already does, since set_size from PixbufLoader was only taken if both values ​​were greater than 0, if it was only one of the two it was ignored.

crates/eww/src/widgets/widget_definitions.rs
hypernova7: Oh I understand now, thanks

w-lfchen (changes requested):
superseded by #1148 now

w-lfchen (changes requested):
can be closed now that #1148 has been merged

PR #1144, opened 2024-07-30
feat: update to 1.80.1 toolchain; dependency updates

w-lfchen:
## Description
**specific details on some crate updates can be found below**
motivated by and closes #1141
closes #1156 
motivated by NixOS/nixpkgs#332957 (eww is currently building in nixpkgs through a lockfile patch)
supersedes #1137 (patch has been applied in a004136779fb0dd0a710d7947158f928e7a8a140)
supersedes #809 (see below), therefore also closes #808
basic functionality tested (systray not tested, my config does not contain one)
reduces build times by dropping ~40 crates in the dependency tree

- update rust toolchain to 1.80.1
- update flake.lock
- update almost all dependencies, 4 have been left out, see below
- switch to nixfmt-rfc-style for nix formatting
- fix new lints
note: there are warnings about unused struct fields now, this probably should be addressed

## `dbusmenu-glib` breaks when automatically updating the missing dependencies
i don't know enough about this to even attempt to fix it, but i'd love to integrate a fix before opening this pr.
### after 4cf9d2d7e460af8f0ff5fcc5378615be95d3ed0e
running `cargo update --dry-run` yields the following:
```
    Updating crates.io index
     Locking 6 packages to latest compatible versions
      Adding gdk-pixbuf-sys v0.20.0
      Adding gio-sys v0.20.0
      Adding glib v0.20.0
      Adding glib-macros v0.20.0
      Adding glib-sys v0.20.0
      Adding gobject-sys v0.20.0
note: pass `--verbose` to see 63 unchanged dependencies behind latest
warning: not updating lockfile due to dry run
```
When adding these crates by simply running `cargo update`, `dbusmenu-glib` fails to compile:
```
error[E0433]: failed to resolve: could not find `ObjectExt` in `glib`
   --> <home>/.cargo/registry/src/index.crates.io-6f17d22bba15001f/dbusmenu-glib-0.1.0/src/auto/client.rs:203:15
    |
203 |         glib::ObjectExt::property(self.as_ref(), "dbus-name")
    |               ^^^^^^^^^ could not find `ObjectExt` in `glib`
    |
help: a struct with a similar name exists
    |
203 |         glib::Object::property(self.as_ref(), "dbus-name")
    |               ~~~~~~
help: consider importing this trait
    |
6   + use glib::prelude::ObjectExt;
    |
help: if you import `ObjectExt`, refer to it directly
    |
203 -         glib::ObjectExt::property(self.as_ref(), "dbus-name")
[...]

PR #1155, opened 2024-08-13
fix: fix changelog, warnings and lints

w-lfchen:
## Description
While working on #1144 I noticed some new warnings when compiling. I'm assuming some warnings were stabilized or added.
This pr fixes these warnings, yielding a slight performance increase, for more details see below.
These are some breaking internal changes since APIs are modified and some call sites might now require generic annotations.

### Replacing the field `display_backend` in the `App` struct with `PhantomData`
This field is never read, producing a warning when compiling with 1.80.0.
It seems to only exist for the type parameter `B`, which has not been restricted to `DisplayBackend` even though it probably should be.
By replacing it with a `PhantomData`, there is no need to pass the `DisplayBackend` around at all times since all the information this argument provides is already encoded in the generic type parameters.

### Removal of field `instance_id` in the `EwwWindow` struct
This field is never read, producing a warning when compiling with 1.80.0.
All information it holds is already in the key of the `open_windows` map in the `App` struct. Removing it also makes the `WindowInitiator`'s `id` field obsolete.

## Checklist

*I don't think this is necessary here*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Fix remove items from systray (By: vnva)
 - Fix the gtk `stack` widget (By: ovalkonia)
 - Fix values in the `EWW_NET` variable (By: mario-kr)
+- Fix the gtk `expander` widget (By: ovalkonia)
 
 ### Features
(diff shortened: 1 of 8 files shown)


elkowar:
image
damn, the way github displays this commit msg confused me for a good second, lmao

Thanks for your cleanup work!

PR #1171, opened 2024-08-26
ci: use clippy; abort run on failure

w-lfchen:
## Description
current state:
- clippy is installed in ci but never invoked
- `cargo check` is executed to check for warnings, but no actionis taken if some exist.

this pr changes both of these things.
clippy is used and if there is a warning or a lint, ci fails.

## Additional Notes
- ci should fail for the first commit, i'll push a fix once that happened (worked as expected).
- it might be worth splitting up things in the workflows so that they can run in parallel.
- i am against ci modifying code.
- i might look into applying some changes from #590

## Checklist
*i don't think this applies here*

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,46 +1,36 @@
 name: build
-
 on:
-   push:
-      branches: [master]
-   pull_request:
-      branches: [master]
-
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
 env:
-   CARGO_TERM_COLOR: always
-   RUSTFLAGS:        "-Dwarnings"
-
+  CARGO_TERM_COLOR: always
+  RUSTFLAGS: "-Dwarnings"
 jobs:
-   build:
-      runs-on: ubuntu-latest
-      steps:
-         - name: Install dependencies
-           run: sudo apt-get update && sudo apt-get install libgtk-3-dev libgtk-layer-shell-dev libdbusmenu-gtk3-dev
-
-         - uses: actions/checkout@v4
-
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -8,6 +8,7 @@ on:
 
 env:
    CARGO_TERM_COLOR: always
+   RUSTFLAGS:        -Dwarnings
```
w-lfchen, line 11: this is needed to fail on warnings, see [the clippy ci docs](https://doc.rust-lang.org/stable/clippy/continuous_integration/github_actions.html).

.github/workflows/build.yml
```diff
@@ -8,6 +8,7 @@ on:
 
 env:
    CARGO_TERM_COLOR: always
+   RUSTFLAGS:        "-Dwarnings"
```
w-lfchen, line 11: this is needed to fail ci, see [the clippy ci docs](https://doc.rust-lang.org/stable/clippy/continuous_integration/github_actions.html)

.github/workflows/build.yml
```diff
@@ -1,45 +1,36 @@
...
-
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
 env:
-   CARGO_TERM_COLOR: always
-
+  CARGO_TERM_COLOR: always
+  RUSTFLAGS: "-Dwarnings"
```
w-lfchen, line 9: this is needed to fail ci, see [the clippy ci docs](https://doc.rust-lang.org/stable/clippy/continuous_integration/github_actions.html)

PR #1174, opened 2024-08-26
Fixing missing commas in example

MunsMan:
Fixing the monitor JSON Example

## Description

In the example, the commas are missing, which lead me to a short hiccup…
Now it is proper JSON and works right out of the box.


w-lfchen (approved):

elkowar:
Damn, that's something I really shouldn't have missed. Thanks a lot!

PR #1176, opened 2024-08-27
feat: support empty string for safe access operator

### LINKED ISSUE #1209: [BUG] Incorrect handling of an empty string by the safe access operator

### Description of the bug

The safe access operator (?.) fails to recognize an empty string on the left side and return null (as it should according to the [documentation](https://elkowar.github.io/eww/expression_language.html#features)), instead throwing an error.

### Reproducing the issue

The way I encountered this issue was using a `deflisten` on `playerctl` output, which outputs a JSON string but produces an empty string when all players are shut down. Here's a minimal `eww.yuck` (for Wayland):
```
(deflisten playerctlstatus 'playerctl metadata -aFf \'{"status": "{{status}}"}\'')

(defwidget playerctl []
    (label :text {playerctlstatus?.status}))

(defwindow test
    :geometry (geometry
                  :width "100%"
                  :height "22px"
                  :anchor "bottom center")
    :exclusive true
    :stacking "fg"
    :monitor 0
    (playerctl))
```
When there is an active player, the label correctly contains its state. However, when the player shuts down, `playerctl` outputs "" and the logs show an error:
```
error: Failed to turn `` into a value of type json-value
  ┌─ /home/mex/.config/eww/test-bug/eww.yuck:4:19
  │
4 │     (label :text {playerctlstatus?.status}))
  │                   ───────────────
```

### Expected behaviour

Either
1. The operator returning a `null` with an empty string on the left, or
2. The [documentation](https://elkowar.github.io/eww/expression_language.html#features) explicitly stating that the operator only works with a `null` on the left side and not an empty string.

### Additional context

_No response_

w-lfchen:
## Description
- Supersedes #629
resolves it's changes requested [here](https://github.com/elkowar/eww/pull/629#issuecomment-1445115831)
- closes #1209

## Additional Notes
This contains the breaking change of no longer supporting safe access for JSON strings (noted in the changelog).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to eww will be listed here, starting at changes since versio
 
 ## Unreleased
 
+### BREAKING CHANGES
+- #1176 changed safe access (`?.`) behavior:
+  Attempting to index in an empty JSON string (`'""'`) is now an error.
+
 ### Fixes
 - Re-enable some scss features (By: w-lfchen)
 - Fix and refactor nix flake (By: w-lfchen)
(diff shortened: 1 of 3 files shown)


w-lfchen (commented):
i dropped the dependency updates, that should be done in a separate pr alongside updating to rust 1.81

elkowar:
I think the breaking change regarding json strings definitely needs to be mentioned in the changelog, but I think conceptually it makes sense this way.

Issue #1165, opened 2024-08-23
[FEATURE] Export magic variables to env in subshell execution


MartyMcFlyInTheSky:
### Description of the requested feature

When executing a script from yuck, e.g.
```

(button :active {loading == false}
                    :class "spinner"
                    :onclick "python scripts/load_networks.py"
                    "${ spinner_symbol }")
```

populate its environment with the [magic variables](https://elkowar.github.io/eww/magic-vars.html) beforehand so the script can make use of them. E.g. in my python script I want to do the following:

```
eww_cmd: str = os.getenv('EWW_CMD')

...

    proc = await asyncio.create_subprocess_shell(f'{eww_cmd} update {vars}',
        stdout=asyncio.subprocess.PIPE,
        stderr=asyncio.subprocess.PIPE)
```

### Proposed configuration syntax

internal

ovalkonia:
Why not just pass all the needed data as command line arguments?

MartyMcFlyInTheSky:
I mean sure, but also why not have them as environmental variables? It would decrease the verbosity and make it more clean imho. As a user I was also quite surprised that this was not a thing, which goes against the principle of least astonishment.

ovalkonia:
Fair enough. As it turns out, there's not way to access variables in `defvar`, `defpoll` and `deflisten`, so there is no way to pass them now. If we were able to do that, we could just pass the needed variables as command line arguments or just set the environmental variables ourselves. Having all magic variables set by default at all times seems kind of excessive, you realistically need one or two. Maybe the access to variables in `defvar` can be made possible

MartyMcFlyInTheSky:
Yes that should be supported at least. Even though I think that exporting 10 variables is not all _that_ excessive, one could export just a subset. I mean for the other variables it might not even make sense to be exported, but `EWW_CMD` has a real use case.

PR #1179, opened 2024-08-29
Export magic constants during server initialization

ovalkonia:
## Description

I was thinking about https://github.com/elkowar/eww/issues/1165 and being able to access the variables within `defvar`, `deflisten` and `defpoll` doesn't really make sense most of the time since they constantly update and there is much trickery involved. So instead why not export the magic constants as environmental variables by default. Makes sense, I suppose

Closes https://github.com/elkowar/eww/issues/1165

## Usage

N/A

## Additional Notes

I also wanted to add some docs regarding this pr and some others where I forgot to do so. Although, `docs/src/magic-vars.md` seems to be almost empty and it doesn't match https://elkowar.github.io/eww/magic-vars.html. I also checked the prs with freshly added features, like `:fill-svg` attribute, but it doesn't change any docs. Still, the website reflects the changes. Am I missing something? Please point me in the right direction

## Checklist

- [] All widgets I've added are correctly documented.
- [] I added my changes to CHANGELOG.md, if appropriate.
- [] The documentation in the `docs/content/main` directory has been adjusted to reflect my changes.

### DIFF
--- a/crates/eww/src/server.rs
+++ b/crates/eww/src/server.rs
@@ -33,7 +33,6 @@ pub fn initialize_server<B: DisplayBackend>(
     log::info!("Loading paths: {}", &paths);
 
     let read_config = config::read_from_eww_paths(&paths);
-
     let eww_config = match read_config {
         Ok(config) => config,
         Err(err) => {
@@ -41,6 +40,10 @@ pub fn initialize_server<B: DisplayBackend>(
             config::EwwConfig::default()
         }
     };
+
+    for (name, definition) in config::inbuilt::get_magic_constants(&paths) {
+        std::env::set_var(name.0, definition.initial_value.0);
+    }
 
     cleanup_log_dir(paths.get_log_dir())?;
 


w-lfchen:
the docs for widgets are automatically generated, but i don't know how magic vars are handled from a glance, it seems to be similar, look at the `@desc` and `@prop` comments

your checkboxes are missing spaces in order to work properly: `- [ ]` is an unchecked one, the space between the brackets is required ^^

some questions i'd like to raise:
- does this export every variable or only the used ones?
- if it exports every variable, is there a noticeable time increase when starting eww? that could be annoying

ovalkonia:
I don't think you can check whether a script uses a certain environmental variable. So it only exports the `magic constants` since they don't change and it only needs to be done once - when the server gets initialized, just like the `GDK_BACKEND` variable.

Therefore, no slowdowns should be present. Right now there are only three `magic constants`: `EWW_CONFIG`, `EWW_CMD` and `EWW_EXECUTABLE` They are defined with the `define_magic_constants` macro. If any more get added in the future, they will be automatically exported as well.

Thank you for the info about the docs. I will mark the pr as draft until I update them

w-lfchen:
i mixed up magic constants and variables, my bad.
this seems reasonable, with little performance overhead then since there's just some strings being thrown around and transformed.

ovalkonia:
Well, I tried my best with the docs (I don't have enough willpower to figure out `gen-docs.ts`)

docs/src/magic-vars.md
```diff
@@ -4,3 +4,6 @@ These are variables that are always there, without you having to import them.
 
 The delay between all the updating variables except `EWW_TIME` is 2s, for `EWW_TIME` it is 1s.
 
+There are also `magic constants`, marked with `(Magic constant)` after the variable's name. As the name implies, they do not change.
+You can also get them from within your scripts since they are automatically exported as environmental variables!
```
w-lfchen, line 8:
```suggestion
You can also access them in your scripts since they are automatically exported as environmental variables!
```

w-lfchen (changes requested):
minor wording, you should be able to simply apply this
apart from that i don't see any issues with this, but i have not tested it.

w-lfchen:
one more typo that i just noticed (would be nice if you could just fix that one as well):
https://github.com/elkowar/eww/blob/b3dbfadfb2143cc0d3e2e21c0cf43ed0a96a53cf/crates/eww/src/config/inbuilt.rs#L33
`Celcius -> degree Celsius`

w-lfchen (changes requested):
while this is a simple export, we should make it useful. `EWW_CMD` currently does not work when invoked as one would expect.

i suspect that the cause for this is the two paths in the command being quoted.

the other variables work and can be used as expected.

also, adding this requires eww to remain single-threaded according to the [`set_env` docs](https://doc.rust-lang.org/std/env/fn.set_var.html). i don't know whether this could cause issues.

w-lfchen:
i phrased that a bit confusingly. `echo` works fine, but consider the following:
`(defpoll test :interval "1s" "$EWW_CMD update foo=bar")`
one would expect this to work (there could be a more complicated script here obviously). however, this is not the case.

ovalkonia:
Got it, that's kinda weird. Without the quotes it indeed works. Although, when I paste the value with quotes manually, it still works. Don't really understand how it works now, but I will look into it

ovalkonia:
So the problem here is that bash interprets quotes literally, as if I was trying to do `'"eww"'` or something like that, which fails of course.

Two other variables are not quoted since you can quote them and escape possible whitespaces yourself. You can't do that with `EWW_CMD`, so it's quoted for you. At least, that's what I think.

What I could do is change the format of `EWW_CMD` to escape whitespaces with `\` instead of quotes. Although, it isn't all that pretty. Is there a better approach?

w-lfchen:
i love bash so much /s
i'll look into an idea i just had

crates/eww/src/config/inbuilt.rs
```diff
@@ -81,17 +81,17 @@ macro_rules! define_magic_constants {
...
 }
 define_magic_constants! { eww_paths,
-    // @desc EWW_CONFIG_DIR - Path to the eww configuration of the current process
+    // @desc EWW_CONFIG_DIR (Magic constant) - Path to the eww configuration of the current process
     "EWW_CONFIG_DIR" => DynVal::from_string(eww_paths.get_config_dir().to_string_lossy().into_owned()),
 
-    // @desc EWW_CMD - eww command running in the current configuration, useful in event handlers. I.e.: `:onclick "${EWW_CMD} update foo=bar"`
+    // @desc EWW_CMD (Magic constant) - eww command running in the current configuration, useful in event handlers. I.e.: `:onclick "${EWW_CMD} update foo=bar"`
     "EWW_CMD" => DynVal::from_string(
         format!("\"{}\" --config \"{}\"",
```
w-lfchen, line 89:
```suggestion
        format!("{} --config {}",
```

w-lfchen (changes requested):
from my admittedly limited testing, this works.
are there issues with this approach?

crates/eww/src/config/inbuilt.rs
ovalkonia: Yeah, that's what I thought, BUT. Try cloning eww into a folder named `eww with   spaces` for example. It doesn't work for me. The path is just cut from the first space to the end. These are kind of things that make want not to deal with bash ever again(

w-lfchen: yeah true, i forgot about that, nvm

hypernova7: Why not use single quotes

ovalkonia: Same problems as with double quotes

ovalkonia:
Since it seems to be a dead end with the environmental variables, thought I would revert those changes and leave just the docs update.

w-lfchen (approved):
i mean sure
i think exporting the two working variables shouldn't do any harm, so we could still export those

ovalkonia:
Yeah, it just kinda doesn't feel right. I really doubt that someone uses `EWW_EXECUTABLE` or `EWW_CONFIG`. `EWW_CMD` makes the most sense, yet it doesn't work. Sure, you can construct `EWW_CMD` using the other two, but meh. So maybe it would be better to come back to this later

elkowar (approved):

PR #1187, opened 2024-09-01
feat: button improvements

w-lfchen:
## Description
- supersedes #640 (that pr can be closed)
- buttons can be activated using the keyboard now
- if buttons play animations on click (catppuccin gtk theme for example), these are now played when they are right-/middleclicked as well
- buttons and eventboxes now consistently activate on key/button release
- some more cleanup in `widget_definitions.rs`

## Usage
currently supported keys:
- space
- return/enter

### Showcase
here's the config i tested with:
```
(defvar left 0)
(defvar mid 0)
(defvar right 0)
(defwindow example
  :monitor 0
  :focusable true
  :geometry (geometry
    :width "100%"
    :height "30px"
    :anchor "top center"
  )
  (button
    :onclick "eww update left=${left+1}"
    :onmiddleclick "eww update mid=${mid+1}"
    :onrightclick "eww update right=${right+1}; eww close example"
    "l: ${left}, m: ${mid}, r: ${right}"
  ))
```

## Additional Notes
we could achieve a bit better performance if we don't want the fancy animation support.
i think the trade-off is worth it, but if you (the person reading this) think differently, feel free to comment. testing is also appreciated

### :warning: help wanted :warning:
i've attempted to implement this logic for eventboxes as well and failed, i don't think this is possible for those, but i'm not sure.
i'd appreaciate it a lot if someone could investigate this further (in a new pr so that this one can be merged)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Add `min` and `max` function calls to simplexpr (By: ovalkonia)
 - Add `flip-x`, `flip-y`, `vertical` options to the graph widget to determine its direction
 - Add `transform-origin-x`/`transform-origin-y` properties to transform widget (By: mario-kr)
+- Add keyboard support for button presses (By: julianschuler)
 
 ## [0.6.0] (21.04.2024)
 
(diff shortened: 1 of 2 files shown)


w-lfchen:
investigating some things

elkowar:
This looks pretty good! We should investigate the event box stuff at some point, but for now this should suffice! Thanks a lot!

PR #1195, opened 2024-09-07
feat: update toolchain to 1.81.0

w-lfchen:
## Description

- switch to 1.81.0 toolchain
- dependency updates
- fix new warnings/lints

## Checklist

*don't think there is anything that needs to be adjusted*

### DIFF
--- a/crates/eww/src/error_handling_ctx.rs
+++ b/crates/eww/src/error_handling_ctx.rs
@@ -31,9 +31,6 @@ pub fn print_error(err: anyhow::Error) {
 }
 
 pub fn format_error(err: &anyhow::Error) -> String {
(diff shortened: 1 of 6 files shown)


crates/eww/src/error_handling_ctx.rs
```diff
@@ -31,9 +31,6 @@ pub fn print_error(err: anyhow::Error) {
 }
 
 pub fn format_error(err: &anyhow::Error) -> String {
-    for err in err.chain() {
-        format!("chain: {}", err);
-    }
```
w-lfchen, line 36: i'm guessing this was some leftover debug print
removing it seemed to make the most sense here

PR #1197, opened 2024-09-16
nix: update flake-compat

w-lfchen:
## Description
update `flake-compat` so that usage is true to [its readme](https://github.com/edolstra/flake-compat?tab=readme-ov-file#usage)

## Additional Notes

i'm not really sure why we need flake-compat at all but i'm no expert so i won't judge this

## Checklist

- [x] I used `nix fmt` to automatically format all code before committing

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,9 +1,6 @@
 {
   inputs = {
-    flake-compat = {
-      url = "github:edolstra/flake-compat";
-      flake = false;
-    };
+    flake-compat.url = "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz";
     nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
     rust-overlay = {
       url = "github:oxalica/rust-overlay";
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -1,9 +1,6 @@
 {
   inputs = {
-    flake-compat = {
-      url = "github:edolstra/flake-compat";
-      flake = false;
-    };
+    flake-compat.url = "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz";
```
w-lfchen, line 3: they recommend using this over the github url. i don't think there is much of a difference in this case since the repo isn't really receiving updates.

flake.nix
```diff
@@ -1,9 +1,6 @@
 {
   inputs = {
-    flake-compat = {
-      url = "github:edolstra/flake-compat";
-      flake = false;
-    };
+    flake-compat.url = "github:edolstra/flake-compat/refs/pull/65/head";
```
w-lfchen, line 3: i don't know whether edolstra/flake-compat#64 affects us, but this patch shouldn't cause harm

PR #1227, opened 2024-11-14
feat: add subcommand to force-poll variable

### LINKED ISSUE #1005: [FEATURE] Manually trigger defpoll variable update

I would like to be able to trigger a `defpoll` variable update so that if a button showing output from a script run periodically gets clicked, it instantly updates.

```
eww update defpoll_variable
```

to update,

```
eww update defpoll_variable=new_walue
```

to force a new value until next update.

### Additional context

I don't want to run all queried scripts every second for obvious resource management reasons. And I'm fine with a 5 second delay in an indicator when updated externally, but when I click on it it feels really jerky and uncomfortable having to wait a few secs before I get any feedback. Still I want the variable to also update on its own (although possible delayed) when updated externally.

tokinanpa:
## Description

This PR adds a subcommand to the CLI to manually poll a polling variable:

```sh
eww poll [NAMES]...
```

This command will force the variable to be polled outside of its interval. It will also ignore `:run-while` conditions.
An error will be thrown if the variable isn't defined with `defpoll` (or if it doesn't exist, obviously).

Resolves #1005

## Additional Notes

This is my first time contributing, and I haven't been able to fully ensure that this code works as intended. Suggestions and bug fixes would be greatly appreciated.

## Checklist

- [x] I added my changes to CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Fix wayland monitor names support (By: dragonnn)
 
 ### Features
+- Add `eww poll` subcommand to force-poll a variable (By: kiana-S)
 - Update rust toolchain to 1.81.0 (By: w-lfchen)
 - Add `:fill-svg` and `:preserve-aspect-ratio` properties to images (By: hypernova7, w-lfchen)
 - Add `:truncate` property to labels, disabled by default (except in cases where truncation would be enabled in version `0.5.0` and before) (By: Rayzeq).
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -45,6 +45,7 @@ use yuck::{
 pub enum DaemonCommand {
     NoOp,
     UpdateVars(Vec<(VarName, DynVal)>),
+    PollVars(Vec<VarName>),
     ReloadConfigAndCss(DaemonResponseSender),
     OpenInspector,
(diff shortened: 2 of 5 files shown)


w-lfchen (commented):
looks good! just two things to note:
1. currently, the command silently ignores non-existent variables, same if no variables are given at all.
judging from `eww update`, this behavior should probably be different.
2. you might want to use [closing keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) in the pr description to link the issue ^^

kiana-S:
>   1. currently, the command silently ignores non-existent variables, same if no variables are given at all.

Strange, I thought I handled the case where the variable doesn't exist. I'll have to look at that again.

kiana-S:
@w-lfchen As far as I can tell, the behavior of `eww update` and `eww poll` are exactly the same here. Neither print an error message to stderr when called with a nonexistent variable, they both put an error message in the logs. The `update` subcommand also does nothing when called with no arguments.
Is there something I'm missing?

w-lfchen:
> Is there something I'm missing?

my bad, i was a bit tired and mixed up `eww update` and `eww get`.
i still think that both shouldn't silently fail but copying `eww update`'s behavior seems like a sensible default.

w-lfchen (approved):
this looks good, thank you for your work! <3

PR #1230, opened 2024-11-18
fix: Load systray items that are registered without a path

Kage-Yami:
## Description

In some environments (e.g. Budgie Desktop on Ultramarine Linux 40, or when using `home-manager`'s `status-notifier-watcher` on Nix), `StatusNotifierItem`s don't have their path as part of their registration, resulting in these items always failing to load in the `systray` widget.

This fixes those "misbehaving" registrations, via DBus introspection; the node tree from the top is crawled until any node implementing `org.kde.StatusNotifierItem` is found, and the path to that node is used as the resulting systray item path.

It also removes string slicing for the same code path that assumed `:`-prefixed addresses without paths must always be a total of 6 characters, when they can be shorter. (This unrelated issue was "hiding" the real issue in some circumstances.)

## Additional Notes

Because DBus introspection (including via `zbus`) spits out XML, `quick-xml` was added as a dependency (and `serde` promoted from a transient dependency to an explicit dependency) to `notifier_host`.

The "crawl" mentioned earlier is implemented by way of a recursive function; given my (albeit naive) understanding of DBus, the trees will never be very deep, so I don't foresee this recursion being problematic. 

## Checklist

- ~All widgets I've added are correctly documented.~
- ~The documentation in the `docs/content/main` directory has been adjusted to reflect my changes.~
  - _(Errm, the `docs/content` directory doesn't seem to exist?)_


Kage-Yami:
Seems this needs a little more work (or at least, may need a "phase 2" to fix more edge-cases); just installed Discord, and it's icon is not showing in the `systray` when using this branch, with the following in the logs:

> `WARN  notifier_host::host > Could not create StatusNotifierItem from address ":1.990": Failure("no StatusNotifierItem found for :1.990")`

I'll investigate this one further...

Kage-Yami:
Ok, what I've got now seems to do the trick. Turns out, for some reason, DBus didn't think any node for Discord implemented the interface (i.e. introspection was empty for every single node, except for listing children).

Interestingly, the previous fallback (that I had removed in favour of the introspection method) of just assuming that it was at `/StatusNotifierItem` would've worked. Here though, I've tweaked that assumption slightly such that if any node in the tree is named exactly `StatusNotifierItem`, then that's probably the one we're after (instead of assuming that it be directly under the root).

elkowar:
Thanks for your work!

PR #1236, opened 2024-11-26
Avoid recreating windows when only `duration` changes

### LINKED ISSUE #260: [FEATURE] When opening an already opened widget it flickers
When a widget window is loaded and eww is invoked to open the same window, there is a tiny flicker of that widget. It is not significant for most use cases, but I sometimes need to run `eww open` periodically. Eww should check if that window is already opened and do nothing to it.

bbb651:
## Description

Avoid recreating windows when only `duration` changes, and instead only updates the window close timer.
This avoids possible flicker with windows such as OSDs. The logic to do this was added in #263 but broke when durations were added.
Fixes #260.  

## Usage

N/A

### Showcase

N/A, but here's my OSD that was flickering (WIP, haven't figured out the scss to remove the margin above the progress):

image

## Additional Notes

N/A

## Checklist

- [x] All widgets I've added are correctly documented. (N/A)
- [x] I added my changes to CHANGELOG.md, if appropriate. (N/A)
- [x] The documentation in the `docs/content/main` directory has been adjusted to reflect my changes. (N/A)

### DIFF
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -367,14 +367,24 @@ impl<B: DisplayBackend> App<B> {
         self.failed_windows.remove(instance_id);
         log::info!("Opening window {} as '{}'", window_args.window_name, instance_id);
 
-        // if an instance of this is already running, close it
-        if self.open_windows.contains_key(instance_id) {
-            self.close_window(instance_id)?;
-        }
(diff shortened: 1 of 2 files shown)


elkowar:
Thanks a lot for your work!

PR #1239, opened 2024-12-01
Render image widgets correctly on HiDPI displays when the display's window scale is not equal to 1.

### LINKED ISSUE #761: [FEATURE] Better image scaling with SVG (on Wayland)

As for now, when I use the image widget with an SVG and an integer scaling on Wayland of 2, they appear blurry. That's probably the case because the Scalable vector graphic is scaled up the same way as any other raster-based image (JPEG, PNG, etc.). The feature I had in mind is to implement a scaling method for any vector-based image without loosing on quality.

It could automatically be detected or explicit stated.

(image :path "icon.svg" :vector true)

### Additional context

I use Hyprland with Wayland and an integer scaling of 2 on a 2880×1440 monitor with HiDPI.

zane-weissman:
## Description

Fixes #761. 

This fix applies to high resolution displays (or any display where the user has set the display scale to something other than 100%). For example, imagine you want to render a 16x16 image (in yuck: `:image-height 16 :image-width 16`) at 2x scale. The resulting image will be 32x32 physical pixels on the display. If your image source has the fidelity (e.g. you have a >32x32 pixel PNG or an SVG), the resulting image should be sharp, but previously, it was being rendered at the _logical_ resolution (16x16) and upscaled to the physical resolution, creating a blurry image.

## Usage

No configuration changes. Images will simply be rendered at high resolution if possible.

### Showcase

Before and after:
Screenshot from 2024-12-01 13-48-26 Screenshot from 2024-12-01 13-54-54

## Additional Notes

All of the practical code is taken from src/widgets/systray.rs. Thank you to whoever wrote that code! Also, I am relatively new to Rust, any comments on style etc. are welcome.

## Checklist

### DIFF
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -13,6 +13,7 @@ use gdk::{ModifierType, NotifyType};
 use glib::translate::FromGlib;
 use gtk::{self, glib, prelude::*, DestDefaults, TargetEntry, TargetList};
 use gtk::{gdk, pango};
+use gtk::{cairo, gdk::ffi::gdk_cairo_surface_create_from_pixbuf};
 use itertools::Itertools;
 use once_cell::sync::Lazy;
 
@@ -591,6 +592,7 @@ fn build_gtk_image(bargs: &mut BuilderArgs) -> Result<gtk::Image> {
         // @prop preserve-aspect-ratio - whether to keep the aspect ratio when resizing an image. Default: true, false doesn't work for all image types
         // @prop fill-svg - sets the color of svg images
         prop(path: as_string, image_width: as_i32 = -1, image_height: as_i32 = -1, preserve_aspect_ratio: as_bool = true, fill_svg: as_string = "") {
+
             if !path.ends_with(".svg") && !fill_svg.is_empty() {
                 log::warn!("Fill attribute ignored, file is not an svg image");
             }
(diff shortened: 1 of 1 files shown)


Kynson:
It would be great if this can get merged, currently facing the same problem. (I know the team maybe busy with other stuff, not intended to rush anyone)

Thanks a lot!

UnnaturalTwilight:
Trying this out on my own fork. Eww fails to load if any images don't have both width and height defined. Most of my config is only defining `:image-width` and using the default `:preserve-aspect-ratio` value of true to implicitly define the height.
I'm using fractional scaling on hyprland.

zane-weissman:
> Trying this out on my own fork. Eww fails to load if any images don't have both width and height defined. Most of my config is only defining `:image-width` and using the default `:preserve-aspect-ratio` value of true to implicitly define the height. I'm using fractional scaling on hyprland.

Thanks for noticing this - I had image_width and image_height both defined in all my configs. It looks like somehow when `image_height` was omitted from the yuck config, it was getting passed into the graphics library as `0` and failing a check when the library tried to scale the image to the requested size. I'm not sure why this happened, because I think the default value for `image_height` should be `-1`... See line 594 of `widget_definitions.rs`:
```rust
prop(path: as_string, image_width: as_i32 = -1, image_height: as_i32 = -1, preserve_aspect_ratio: as_bool = true, fill_svg: as_string = "") {
``` 

Anyways, this fix just sets the `image_height` and/or `image_width` to `-1` if they're not positive before passing them to `gtk::gdk_pixbuf::Pixbuf::from_file_at_scale()`. Let me know if it's working for you too!

Sempyos:
Trying this out on my end works wonderfully on my HiDPI display. I will note that it also seems css background images also have the same issue as widgets
image

PR #1217, opened 2024-10-20
feat: update to 1.82.0 toolchain; dependency updates

w-lfchen:
## Description

- the flake fell victim to NixOS/nixpkgs#348832 due to the rust overlay being not up to date, potentially leading to evaluation failures when overriding the `nixpkgs` input which is fixed by the lockfile update in https://github.com/elkowar/eww/commit/89c372d128d0c34a28bf06f890e898385a17cb49.

## Checklist

- [x] I added my changes to CHANGELOG.md, if appropriate.
- [x] I used `cargo fmt` to automatically format all code before committing
- [x] tested compilation, tests and clippy

PR #1248, opened 2024-12-17
Add :escape property to label

silicon-salamander:
## Description

Add `:escape` property to the label widget in order to disable interpreting escape sequences. Defaults to true.

## Usage

```yuck
(label :class "window-title"
  :escape false
  :text {title} ; could contain any text that I don't want interpreted
)
```

### Showcase

With `:escape true` the window title on my bar shows newlines and expands the bar.
screenshot_2024-12-17_12:12:51

With `:escape false` the window title correctly displays `\n` as literal text.
screenshot_2024-12-17_12:12:25

## Additional Notes

I'm not sure if I need to add documentation anywhere else. Let me know if I do.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Add `transform-origin-x`/`transform-origin-y` properties to transform widget (By: mario-kr)
 - Add keyboard support for button presses (By: julianschuler)
 - Support empty string for safe access operator (By: ModProg)
+- Add `:escape` property to label. Determines weather escape sequences are interpreted in the `:text` property. (By: silicon-salamander)
 
 ## [0.6.0] (21.04.2024)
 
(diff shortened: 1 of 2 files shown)


w-lfchen (commented):
i think a good thing to add to this pr would be switching to a maintained escaping crate such as [unescaper](https://crates.io/crates/unescaper) (haven't really looked for more).

i started working on improving the escaping logic a while ago since in some cases, things need to be escaped twice, which shouldn't be necessary, but these days i don't really have enough free time to finish my draft.

if you want to, you can look into this as well, but it's probably out of scope for this pr

silicon-salamander:
Hmm yeah, I didn't realize there were wider issues. I guess one of the lexers needs to be modified to work better in combination with `unescape` so that you don't need to do `\\n`. 

I think the ideal fix would be to only have string literals in your config file automatically "unescaped" and then have an `escape()` function in simplexpr so you can optionally "unescape" text that is coming in through a dynamic variable.

This PR is just a bandage, but not really fixing the root cause. I barely know any rust, so I would need to learn more to be able to fix these issues. I'm definitely interested in trying, it just might take a while.

w-lfchen:
i think your diagnosis is correct, that's the conclusion i came to as well after a bit of testing. i think there was a page in the lalrpop docs which was relevant, but it's been a while so i might be mistaken.
imo this api is good to have either way, so id love to see this added.

i'll add the dependency change to #1217 since `unescaper` seems to be a direct replacement for `unescape`, but this might lead to merge conflicts if it is merged before this one

Issue #1226, opened 2024-11-08
[FEATURE] "pow" binary operation in simpleexpr


elisesouche:
### Description of the requested feature

Simple mathematical expressions can be evaluated in eww. These include +, *, and others. However, there is no support for exponentiation. I needed it for manipulating tags in RiverWM, but I have to shell out.

### Proposed configuration syntax

We could use either the syntax `a ** b` as in Python or the more common `a ^ b`.

### Additional context

I can probably make a PR for this if you give me a green flag.

PR #1255, opened 2024-12-23
`powi`/`powf` simplexpr functions

ovalkonia:
## Description

Yep, `powi`/`powf`, why not

Resolves https://github.com/elkowar/eww/issues/1226

## Usage

In the showcase

### Showcase

23-12-24 at 23:08:48

## Additional Notes

Sooo, it's two separate expressions. They can be reduced down to a single one, that acts like `powf`. I just though it sometimes miiight be useful to have two of them. Just shrink it down to one expression if that's more appropriate

## Checklist

### DIFF
--- a/crates/simplexpr/src/eval.rs
+++ b/crates/simplexpr/src/eval.rs
@@ -345,6 +345,22 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
             }
             _ => Err(EvalError::WrongArgCount(name.to_string())),
         },
+        "powi" => match args.as_slice() {
+            [num, n] => {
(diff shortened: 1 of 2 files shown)


w-lfchen (approved):
lgtm, but maybe consider adding this to the grammar using operators as described in #1226

ovalkonia:
> consider adding this to the grammar using operators as described in #1226

I kinda like it more in the form of a function. I mean, if it's more appropriate to implement it the other way, I would. Although, I would prefer for it to be this way

elkowar:
This seems fine to me, an operator would also work but I don't really see a massive benefit either way. I'll merge this! Thanks for your contribution!

Issue #1158, opened 2024-08-16
[BUG] Crash when monitor turned off


tsvallender:
### Checklist before submitting an issue

- [X] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [X] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [X] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

### Description of the bug

Turning off the monitor causes eww to crash.

### Reproducing the issue

If eww is running and the monitor is turned off, when turned back on eww will have crashed.

From the logs:
```
 2024-08-16T08:13:31.265Z INFO  eww::app        > Closing gtk window bar

(eww:86616): GLib-GObject-CRITICAL **: 09:13:31.265: ../gobject/gsignal.c:2685: instance '0x564390eb3d80' has no handler with id '92'
 2024-08-16T08:13:31.265Z ERROR eww::error_handling_ctx > Failed to send success response from application thread

Caused by:
    channel closed
```

This seems unrelated to config, I’ve removed my config and replaced it with the barebones config below and still have the issue.

Running eww 0.6.0 on openSUSE.

Config:

```
(defwidget bar [] (music))
(defwindow bar
  :monitor 0
  :windowtype "dock"
  :exclusive true
  :geometry (geometry :x "0%"
                      :y "0%"
                      :width "90%"
                      :height "10px"
                      :anchor "top center")
  :reserve (struts :side "top" :distance "100px")
  (bar))

(defwidget music []
  (box :class "music"
       :orientation "h"
       :space-evenly false
       :halign "center"
    {"hello"}))
```

cstruck:
can confirm. whenever I turn off my monitor the eww window crashes

elkowar:
is this happening on wayland or on X11? Is it actually related to the monitor, or is it the entire computer going to sleep?
If wayland, we might need to do some explicit work to catch this sort of thing, Idk exactly how compositors deal with monitors vanishing and popping up

cstruck:
oh sry, should have added that. Yeah, its wayland.

it crashes with

```
[...]

PR #1267, opened 2025-01-11
Add disk io stats to EWW_DISK

bkueng:
Very cool project.

## Description

This updates system_stats and exposes `read` and `written` to `EWW_DISK`, allowing to monitor disk io stats.

## Checklist

Please make sure you can check all the boxes that apply to this PR.

- [ ] All widgets I've added are correctly documented.
- [ ] I added my changes to CHANGELOG.md, if appropriate.
- [ ] The documentation in the `docs/content/main` directory has been adjusted to reflect my changes.
- [ ] I used `cargo fmt` to automatically format all code before committing

PR #1276, opened 2025-01-25
Improve multi-monitor handling under wayland

bkueng:
When a monitor gets disconnected, the destroy event of all associated windows gets called, and the window gets removed.

This PR changes that behavior: the window is still closed but the configuration is kept using the existing reload mechanism. In addition, a callback is added to listen for new monitors, triggering a reload when a new monitor gets connected.

This logic also reloads already running windows, which has a positive and negative effect:
- positive: if currently running e.g. on the second monitor specified in the list, the window can get moved to the first monitor
- negative: if reloading starts it on the same monitor, it gets reset (e.g. graphs)

I also had to work around an issue: the monitor model is not yet available immediately when a new monitor callback triggers. Waiting in the callback does not help (I tried 10 seconds). However, waiting outside, it always became available after 10ms.

Tested with a dual monitor setup under KDE through a combinations of:
- enabling/disabling individual monitors
- switching between monitors
- specifying a specific monitor in the yuck config
- specifying a list of specific monitors in the yuck config

In all these cases the behavior is as expected, and the widget gets loaded on the first available monitor (or stays unloaded until one becomes available).
It also works when opening a window without any of the configured monitors being available.

There is one remaining error from GTK when closing the window: `GLib-GObject-CRITICAL **: 20:06:05.912: ../gobject/gsignal.c:2684: instance '0x55a4ab4be2d0' has no handler with id '136'`.
 This comes from the `self.gtk_window.disconnect(handler_id)` call. To prevent that we'd have to reset `destroy_event_handler_id`.

I have not tested under X11, and I'm not sure if the behavior should be the same there.

Fixes https://github.com/elkowar/eww/issues/1158

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Fix wayland monitor names support (By: dragonnn)
 - Load systray items that are registered without a path (By: Kage-Yami)
 - `get_locale` now follows POSIX standard for locale selection (By: mirhahn, w-lfchen)
+- Improve multi-monitor handling under wayland (By: bkueng)
 
 ### Features
 - Add warning and docs for incompatible `:anchor` and `:exclusive` options
--- a/crates/eww/src/app.rs
+++ b/crates/eww/src/app.rs
@@ -68,6 +68,7 @@ pub enum DaemonCommand {
     },
     CloseWindows {
         windows: Vec<String>,
+        auto_reopen: bool,
         sender: DaemonResponseSender,
     },
     KillServer,
@@ -147,16 +148,35 @@ impl<B: DisplayBackend> std::fmt::Debug for App<B> {
     }
 }
 
+/// Wait until the .model() is available for all monitors (or there is a timeout)
+async fn wait_for_monitor_model() {
+    let display = gdk::Display::default().expect("could not get default display");
+    let start = std::time::Instant::now();
+    loop {
+        while gtk::events_pending() && !gtk::main_iteration_do(false) {}
+        let all_monitors_set =
+            (0..display.n_monitors()).all(|i| display.monitor(i).and_then(|monitor| monitor.model()).is_some());
+        if all_monitors_set {
+            break;
+        }
+        tokio::time::sleep(Duration::from_millis(10)).await;
+        if std::time::Instant::now() - start > Duration::from_millis(500) {
+            log::warn!("Timed out waiting for monitor model to be set");
+            break;
+        }
+    }
+}
+
 impl<B: DisplayBackend> App<B> {
     /// Handle a [`DaemonCommand`] event, logging any errors that occur.
-    pub fn handle_command(&mut self, event: DaemonCommand) {
-        if let Err(err) = self.try_handle_command(event) {
+    pub async fn handle_command(&mut self, event: DaemonCommand) {
+        if let Err(err) = self.try_handle_command(event).await {
             error_handling_ctx::print_error(err);
         }
     }
 
     /// Try to handle a [`DaemonCommand`] event.
-    fn try_handle_command(&mut self, event: DaemonCommand) -> Result<()> {
+    async fn try_handle_command(&mut self, event: DaemonCommand) -> Result<()> {
         log::debug!("Handling event: {:?}", &event);
         match event {
             DaemonCommand::NoOp => {}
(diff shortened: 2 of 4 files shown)


elkowar:
I'm down to merge this once someone has tested if the behavior breaks anything under X11 ^^

bkueng:
I still have an old system with X11 installed, and was able to test there. It works as well with 2 differences in behavior:
- the 'on new monitor' callback gets called twice whenever a new monitor is connected
- the window does not get distroyed when the underlying monitor is disconnected.

I also rebased and resolved the conflicts.

There was a regression in a recently added commit (https://github.com/elkowar/eww/commit/29fa1587c3a4704a2bee0778099aca945d486e77) which I fixed in https://github.com/elkowar/eww/pull/1276/commits/6b0f42e24610d6b1cbe1bde2685e59a501c22f09.

bkueng:
@elkowar is there anything else I can or should do to get this merged?

Flat:
I've been using this branch for a few days now, and when I re-connect monitors it causes a SIGABRT. Backtrace attached. Very consistently happens every time. 
[gdb.txt](https://github.com/user-attachments/files/19234166/gdb.txt)

bkueng:
@Flat which Linux distribution and desktop environment / window manager do you use?
And could you provide the yuck config, maybe it helps to reproduce?

Flat:
> @Flat which Linux distribution and desktop environment / window manager do you use? And could you provide the yuck config, maybe it helps to reproduce?

Arch Linux with Hyprland (From git master)

Attached is the yuck config as well.
[eww.txt](https://github.com/user-attachments/files/19297854/eww.txt)

From the crash dump, it appears to be coming from this https://github.com/elkowar/eww/pull/1276/files#diff-12f72e340b5f9de88e64ebfc3df46f5d14a2bc5b0b2221eadf43ba087765d762R157-R158 as it is the only closure in wait_for_monitor_model

I would assume it is hitting something on the monitors before they are fully initialized.

cstruck:
It often works but sometimes eww crashes

```
thread 'main' panicked at /var/tmp/portage/gui-apps/eww-9999/work/cargo_home/gentoo/glib/src/main_context_futures.rs:238:56:
called `Result::unwrap()` on an `Err` value: EnterError
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at library/core/src/panicking.rs:218:5:
panic in a function that cannot unwind
stack backtrace:
   0:     0x55a4fdb5ceba - <unknown>
   1:     0x55a4fdb89ed3 - <unknown>
   2:     0x55a4fdb57dd3 - <unknown>
   3:     0x55a4fdb5cd02 - <unknown>
   4:     0x55a4fdb5deb0 - <unknown>
   5:     0x55a4fdb5dc90 - <unknown>
   6:     0x55a4fdb5e612 - <unknown>
   7:     0x55a4fdb5e386 - <unknown>
   8:     0x55a4fdb5d3b9 - <unknown>
   9:     0x55a4fdb5e04d - <unknown>
  10:     0x55a4fd03b36d - <unknown>
  11:     0x55a4fd03b402 - <unknown>
  12:     0x55a4fd03b526 - <unknown>
  13:     0x55a4fdb28127 - <unknown>
  14:     0x7fafa0810ae2 - <unknown>
  15:     0x7fafa0813f07 - <unknown>
  16:     0x7fafa08146b0 - g_main_context_iteration
  17:     0x7fafa0ff73f5 - gtk_main_iteration_do
  18:     0x55a4fdb202eb - <unknown>
  19:     0x55a4fd1a236b - <unknown>
  20:     0x55a4fd1a5a3d - <unknown>
  21:     0x55a4fd18e35e - <unknown>
  22:     0x55a4fdb27e18 - <unknown>
  23:     0x7fafa0810ae2 - <unknown>
  24:     0x7fafa0813f07 - <unknown>
  25:     0x7fafa081499f - g_main_loop_run
  26:     0x7fafa0ff72fd - gtk_main
  27:     0x55a4fd2443f9 - <unknown>
  28:     0x55a4fd24a09f - <unknown>
  29:     0x55a4fd1289d3 - <unknown>
  30:     0x55a4fd3178e9 - <unknown>
  31:     0x55a4fdb4db07 - <unknown>
  32:     0x55a4fd24ce05 - <unknown>
  33:     0x7fafa05201ee - <unknown>
  34:     0x7fafa05202a9 - __libc_start_main
  35:     0x55a4fd03ba35 - <unknown>
  36:                0x0 - <unknown>
thread caused non-unwinding panic. aborting.
```

bkueng:
Thanks for the data points. I was able to find the problem and pushed a fix.
Could you give it another try?

Flat:
> Thanks for the data points. I was able to find the problem and pushed a fix. Could you give it another try?

So far looks like it is now working as intended!

Philipp-M:
I've tested it on both X as well as wayland where the bug indeed seems to be fixed.
Thanks for fixing this!

cstruck:
can confirm didn't happen since

Sahnvour:
That would really help for my setup, using a kvm switch means I need to restart eww each time I switch video input.

cstruck:
@elkowar anything you still need to get this merged?

bkueng:
@elkowar I understand the problem of having little time but I do hope it's the last time I rebased. I also have 2 other PR's which have conflicts by now: https://github.com/elkowar/eww/pull/1267 and https://github.com/elkowar/eww/pull/1277

PR #1277, opened 2025-01-25
feat: add byteshumanreadable(bytes, num_digits) to expressions

bkueng:
## Description

Can be used to display bytes, e.g. free disk space `${byteshumanreadable(EWW_DISK["/"].free,1)}` -> `150.3 GB`.

Or also for rates, e.g. `${byteshumanreadable(EWW_NET.wlp3s0.NET_DOWN,0)}/s` -> `392 KB/s`

The name is a bit long, but I could not think of anything better.

## Checklist

### DIFF
--- a/crates/simplexpr/src/eval.rs
+++ b/crates/simplexpr/src/eval.rs
@@ -508,6 +508,33 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
             }
             _ => Err(EvalError::WrongArgCount(name.to_string())),
         },
+        "byteshumanreadable" => match args.as_slice() {
+            [size, digits] => {
+                let mut size = size.as_f64()?;
+                let digits = if size > 0.0 { digits.as_i32()? } else { 0 };
+                let mut suffix = "B";
+                while size >= 1024.0 {
+                    match suffix {
+                        "B" => {
+                            suffix = "KB";
+                        }
+                        "KB" => {
+                            suffix = "MB";
+                        }
+                        "MB" => {
+                            suffix = "GB";
+                        }
+                        "GB" => {
+                            suffix = "TB";
+                        }
+                        _ => break,
+                    }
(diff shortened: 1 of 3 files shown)


crates/simplexpr/src/eval.rs
```diff
@@ -508,6 +508,33 @@ fn call_expr_function(name: &str, args: Vec<DynVal>) -> Result<DynVal, EvalError
             }
             _ => Err(EvalError::WrongArgCount(name.to_string())),
         },
+        "byteshumanreadable" => match args.as_slice() {
+            [size, digits] => {
+                let mut size = size.as_f64()?;
+                let digits = if size > 0.0 { digits.as_i32()? } else { 0 };
+                let mut suffix = "B";
+                while size >= 1024.0 {
```
DOD-101, line 516: If this is added, it would be good to allow users to choose between binary and decimal prefixes: https://en.wikipedia.org/wiki/Byte#Multiple-byte_units and adjust the code accordingly.

bkueng: I can add that. Maybe with an optional argument `binary`/`decimal`?

crates/simplexpr/src/eval.rs
```diff
@@ -548,6 +568,31 @@ fn run_jaq_function(json: serde_json::Value, code: String, args: &str) -> Result
...
 }
 
+fn bytes_human_readable(size: f64, digits: i32, unit_is_binary: bool) -> String {
+    let mut size = size;
+    let mut suffix = "B";
+    let base = if unit_is_binary { 1024.0 } else { 1000.0 };
+    while size >= base {
+        match suffix {
+            "B" => {
+                suffix = "KB";
```
DOD-101, line 578: The suffixes should also be adjusted accordingly. 

Also, it might be better to not do this using a loop (just thinking in terms of performance). I would use an ordered array of suffixes and just index into that.

PR #1278, opened 2025-01-27
add lines and wrap-mode properties to label widget

vaporii:
## Description

Add the `:lines` and `:wrap-mode` properties to the label widget. Used for setting the number of lines to display when wrapping is enabled, and how it should display wrapped text, respectively.

## Usage

To set the `lines` and `wrap-mode` properties, you could do something like this:
```
(label
  :text "this is some label text blah blkaghk kwjhndlkaj wbn akj hba jkhawbjakhjhawjkdbh jahbw ajkhbakjawja"
  :wrap true
  :limit-width 15
  :lines 4 ; the number of lines to display
  :wrap-mode "char" ; wraps on characters, ignoring words
)
```
This will look like the first image in the showcase below.

### Showcase

line-wrap-example

## Additional Notes

When the label text contains newlines, they aren't interpreted as lines. This fix involves changing the implementation of newlines in labels, or implementing some janky fix elsewhere

## Checklist

### DIFF
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -970,7 +970,8 @@ fn build_gtk_label(bargs: &mut BuilderArgs) -> Result<gtk::Label> {
         // @prop truncate-left - whether to truncate on the left side
         // @prop show-truncated - show whether the text was truncated. Disabling it will also disable dynamic truncation (the labels won't be truncated more than `limit-width`, even if there is not enough space for them), and will completly disable truncation on pango markup.
         // @prop unindent - whether to remove leading spaces
-        prop(text: as_string, truncate: as_bool = false, limit_width: as_i32 = i32::MAX, truncate_left: as_bool = false, show_truncated: as_bool = true, unindent: as_bool = true) {
+        // @prop lines - maximum number of lines to display (only works when `limit-width` has a value)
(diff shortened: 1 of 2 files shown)


crates/eww/src/widgets/widget_definitions.rs
```diff
@@ -1050,6 +1050,14 @@ fn build_gtk_label(bargs: &mut BuilderArgs) -> Result<gtk::Label> {
         prop(justify: as_string = "left") {
             gtk_widget.set_justify(parse_justification(&justify)?);
         },
+        // @prop wrap-mode - how text is wrapped. possible options: $wrap-mode
+        prop(wrap_mode: as_string = "word") {
+            gtk_widget.set_wrap_mode(parse_wrap_mode(&wrap_mode)?);
+        },
+        // @prop lines - maximum number of lines to display (only works when `limit-width` has a value)
+        prop(lines: as_i32 = -1) {
```
w-lfchen, line 1058: i'm guessing that -1 disables the limit, which should be documented imo

w-lfchen (commented):
looks good overall, thank you for your work <3

crates/eww/src/widgets/widget_definitions.rs
vaporii: fixed!

w-lfchen (approved):
lgtm, thank you again <3

PR #1315, opened 2025-05-08
Add floor() and ceil()

wsbankenstein:
## Description

Add `floor()` and `ceil()` rounding functions to simplexpr.
No additional dependencies or code changes.

## Checklist

- [V] I added my changes to CHANGELOG.md, if appropriate.
- [V] The documentation in the `docs/content/main` directory has been adjusted to reflect my changes.
- [V] I used `cargo fmt` to automatically format all code before committing


elkowar:
Your current docs suggest that the floor and ceil functions take two arguments, whereas your implementation seems to only take the actual number (no decimal-digit count). I'd suggest we add both cases, so it defaults to flooring to 0 decimal points, but if an additional argument is provided, those get considered as well

wsbankenstein:
Yeah, I removed the extra argument but forgot to update the documentation. I don't think an optional second argument is needed, because that functionality can be achieved easily with `round(floor(num), digits)`. Also, it's not the most necessary functionality to have, seeing as all the digits after the decimal point would be zeroes.

Documentation is now updated.

Issue #1290, opened 2025-03-01
[BUG] Wrong format inside "formattime" results in the whole program being unusable


RQuarx:
### Checklist before submitting an issue

- [x] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [x] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [x] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

### Description of the bug

Inserting any other characters that is not the format specifier of the formattime would result in the whole program being unusable / needing a full restart to wrok

### Reproducing the issue

formattime(EWW_TIME, "%EWW")

### Expected behaviour

EWW would stop responding to everything until the formattime is fixed, and a bunch of errors were spit out on the log

### Additional context

![Image](https://github.com/user-attachments/assets/840c68bd-d711-4862-9ccf-2fb9515a34ab)

![Image](https://github.com/user-attachments/assets/ecaff319-4d94-4d16-8e76-423e7b531400)

### Platform and environment

- compositor: Hyprland (Wayland)
- distro: ArchLinux
- eww_version: 0.5.0 d87c2fdbfdc012e76d229e4e9ea3325bc0f23e89

PR #1317, opened 2025-05-11
Fix crash on invalid formattime

luca3s:
## Description

Fixes https://github.com/elkowar/eww/issues/1290, by checking for an error when formatting with chrono.
Edit: Also updates chrono to use the [write_to](https://docs.rs/chrono/latest/chrono/format/struct.DelayedFormat.html#method.write_to) API.

## Additional Notes

~~If you would like i could also update chrono, which would allow to use the [write_to](https://docs.rs/chrono/latest/chrono/format/struct.DelayedFormat.html#method.write_to) method instead of the rust write! macro. This would probably be faster and it would be more obvious why there could be an error. This also would remove the need to import `fmt::Write`~~
Maybe this could be made into a function, but because the Chrono time objects are generic with different types of time this isn't obvious how to do, so i just didn't.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to eww will be listed here, starting at changes since versio
   Attempting to index in an empty JSON string (`'""'`) is now an error.
 
 ### Fixes
+- Fix crash on invalid `formattime` format string (By: luca3s)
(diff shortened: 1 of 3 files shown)


elkowar:
updating chrono sounds quite nice here, go for it!

PR #1360, opened 2025-07-23
Proper escaping for string literals

netfri25:
## Description

Applies unescaping for string literals in the lexer.
No things have been added to implement this, since it makes use of the `unescape` crate that was used in a different place.

## Usage

"next line\ntabbed\tspace using hex code\x20"

## Additional Notes

I'm not sure if this is a fix or a feature, so for now I haven't added anything to `CHANGELOG.md`

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2711,6 +2711,7 @@ dependencies = [
  "strsim",
  "strum",
  "thiserror",
+ "unescape",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


w-lfchen (commented):
i feel like this pr should include switching the escaping crate, see #1194.
i might pick that pr up at some point with an actual implementation but i've lost the energy, time and motivation to work on eww as of writing.
just thought i should point this out because it seemed relevant

netfri25:
this wasn't hard to implement, so I've just done it myself.
for now it doesn't really handle any errors, and when an error occures when escaping a string literal then it just returns the original unescaped string.

PR #1401, opened 2025-10-30
feat: use lockfiles to determine if daemon is running

### LINKED ISSUE #255: [BUG]  Eww spawns zombie windows and processes
## Checklist before submitting an issue
- [x] I have searched through the existing [closed and open issues](https://github.com/elkowar/eww/issues?q=is%3Aissue) for eww and made sure this is not a duplicate
- [x] I have specifically verified that this bug is not a common [user error](https://github.com/elkowar/eww/issues?q=is%3Aissue+label%3Ano-actual-bug+is%3Aclosed)
- [x] I am providing as much relevant information as I am able to in this bug report (Minimal config to reproduce the issue for example, if applicable)

## Describe the bug
Basically, launching widgets will open your widgets, but sometimes they'll spawn, but refuse to close. Trying `eww kill` will kill the daemon, but these windows will not be closed. `killall eww`, however, works.

As far as I can predict, this issue happens after making `eww daemon` not a dependency for launching eww widgets. I'm not quite sure though.

## Reproducing the issue
Use any config and create a widget. Bind the opening of your window to a keybind. Do not launch `eww daemon`. Press your keybind. Eww will not open windows immediately which is expected. 

Now, before eww has finished opening the window, press the keybind again. Try to close this window with the `eww close` command. It will not work. The only way to remove the window will be `killall eww`.

This also happens sometimes randomly even when the daemon is running, but I cannot consistently reproduce it.

## Expected behaviour
Eww windows to close with the `eww close` or `eww kill` commands.

## Additional context
Logs:
image

The log says `Closing all windows`. However, windows aren't closed.

paperluigis:
## Description
This PR makes the daemon create and lock a file next to the IPC socket (ex. "/run/user/1000/eww-server_f12a67507f690016.lock"). It also makes the program check if the daemon is running by seeing if the file is locked by another process, instead of sending a ping command to the daemon.

This will fix #255, where a timeout in sending an `open` command over IPC caused the program to create a new daemon, leading to "zombie windows" being left over by the existing daemon.
This will also hopefully make the daemon launch a bit quicker when started as `eww open bar`.

## Additional Notes
i haven't tested this on platforms other than linux

is this a feature or a fix?

this uses the `std::fs::File::try_lock` method which is cleaner but only available in rust 1.89.0 and above; i have an alternative implementation that uses `nix::fcntl::Flock` instead, should i submit the alternative instead?

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to eww will be listed here, starting at changes since versio
 - Improve multi-monitor handling under wayland (By: bkueng)
 
 ### Features
+- Use a lock file to determine if a daemon is already running, instead of pinging (By: paperluigis)
 - Add warning and docs for incompatible `:anchor` and `:exclusive` options
 - Add `eww poll` subcommand to force-poll a variable (By: kiana-S)
 - Add OnDemand support for focusable on wayland (By: GallowsDove)
--- a/crates/eww/src/main.rs
+++ b/crates/eww/src/main.rs
@@ -119,20 +119,20 @@ fn run<B: DisplayBackend>(opts: opts::Opt, eww_binary_name: String) -> Result<()
             false
         }
 
-        // make sure that there isn't already a Eww daemon running.
-        opts::Action::Daemon if check_server_running(paths.get_ipc_socket_file()) => {
-            eprintln!("Eww server already running.");
-            true
-        }
         opts::Action::Daemon => {
-            log::info!("Initializing Eww server. ({})", paths.get_ipc_socket_file().display());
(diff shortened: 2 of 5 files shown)


w-lfchen (commented):
since eww has a `rust-toolchain.toml` file that currently is set to 1.81, this pr will break anything relying on that file to determine the rust version to compile eww with due to using newer apis.
in my opinion, the right move here would be updating said file and fixing issues that may arise in the process.
i currently don't really have time to do so for personal reasons, would love to do it otherwise.

w-lfchen:
there could be new warnings or lints

PR #1452, opened 2026-07-25
Fix :visible updates being lost for unmapped widgets

### LINKED ISSUE #1451: `:visible` updates are lost for widgets inside a closed revealer (stale value applied at first map)

- [x] I have searched through the existing closed and open issues for eww and made sure this is not a duplicate
- [x] I have specifically verified that this bug is not a common user error
- [x] I am providing as much relevant information as I am able to in this bug report

### Description of the bug

`:visible` updates issued while a widget is **not yet mapped** — most commonly a child of a closed `revealer` — are silently overwritten when the widget finally maps. The initial `:visible` value is evaluated at build time and re-applied at the widget's *first map* (`connect_first_map` in `widget_definitions.rs`, needed to survive the window's `show_all`). But that first map can happen long after build, and the captured snapshot clobbers every reactive update made in between.

Concrete symptom: with `(defvar vis false)`, setting `vis=true` while the revealer is closed, then revealing → the widget stays **hidden** even though `eww state` says `vis: true`. Toggling `vis` again *while revealed* works fine (the stale one-shot map handler has been consumed by then), which makes the bug look intermittent.

The reverse order (hide while closed) happens to behave correctly, because a hidden widget never maps, so the stale handler stays dormant — the asymmetry is what makes this so confusing to debug in a real config.

### Reproducing the issue

```
; eww.yuck
(defvar reveal false)
(defvar vis false)

(defwindow repro
  :monitor 0
  :stacking "overlay"
  :geometry (geometry :x "40px" :y "80px" :width "280px" :height "170px" :anchor "top left")
  (box :orientation "v" :space-evenly false
    (label :text "reveal: ${reveal} / vis: ${vis}")
    (revealer :reveal reveal :transition "slidedown" :duration "150ms"
      (box :orientation "v" :space-evenly false
        (label :text "inside revealer:")
        (label :visible vis :text ">>> TOGGLE-ME <<<")))))
```

```console
$ eww --config <dir> daemon --no-daemonize &
$ eww --config <dir> open repro
$ eww --config <dir> update vis=true      # while the revealer is closed
$ eww --config <dir> update reveal=true   # TOGGLE-ME should appear — it does not
```

Reproduced on the v0.6.0 release binary and on current master (48f5aa8):

| after `vis=true` + `reveal=true` (master) |
| --- |
| before |

The header says `vis: true` but `>>> TOGGLE-ME <<<` is missing.

### Expected behaviour

The widget reflects the *current* value of `:visible` when it becomes visible/mapped — same as if the update had been issued while revealed:

| same sequence with the fix |
| --- |
| after |

### Additional context

Root cause is the build-time capture in `resolve_widget_attrs` (`crates/eww/src/widgets/widget_definitions.rs`): the first-map handler applies a `visible` value snapshotted at build. I have a small fix (share the latest value in a cell that reactive updates write to) with before/after verification — PR incoming.

61021:
## Description

Fixes #1451.

The initial value of `:visible` is applied at the widget's **first map** (`connect_first_map`) so that it survives the window's `show_all` — but the value was captured at build time. For widgets whose first map happens later (children of a closed `revealer` being the common case), every reactive `:visible` update issued in between was clobbered by that stale snapshot the moment the widget finally mapped: `update vis=true` while the revealer was closed left the widget hidden after revealing, even though `eww state` showed the right value.

The fix keeps the latest value in a shared `Rc<Cell<bool>>`: the reactive `visible` prop writes to it, and the first-map handler reads it instead of a build-time copy. No behavior change for the ordinary cases (initial value application and reactive toggling while mapped work exactly as before — the cell starts with the evaluated initial value).

## Usage

No configuration changes — existing configs just start behaving as expected:

```lisp
(defvar reveal false)
(defvar vis false)

(revealer :reveal reveal
  (label :visible vis :text "hello"))
```

```console
$ eww update vis=true     ; while the revealer is closed
$ eww update reveal=true  ; label now correctly appears
```

### Showcase

Same sequence (`vis=true` while closed → reveal), before and after, on master 48f5aa8:

| before (bug: `vis: true` but widget hidden) | after |
| --- | --- |
| before | after |

The hide-while-closed direction (which already worked, since a hidden widget never maps) is unchanged:

hide direction

## Additional Notes

- Verified on a live Wayland/Hyprland session (both directions, plus toggling while revealed).
- `cargo fmt` and `cargo clippy` clean; CHANGELOG entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to eww will be listed here, starting at changes since versio
   Attempting to index in an empty JSON string (`'""'`) is now an error.
 
 ### Fixes
+- Fix `:visible` updates being lost for widgets that have not yet been mapped, e.g. inside a closed revealer (By: 61021)
 - Fix crash on invalid `formattime` format string (By: luca3s)
 - Fix crash on NaN or infinite graph value (By: luca3s)
 - Re-enable some scss features (By: w-lfchen)
--- a/crates/eww/src/widgets/widget_definitions.rs
+++ b/crates/eww/src/widgets/widget_definitions.rs
@@ -17,7 +17,7 @@ use itertools::Itertools;
 use once_cell::sync::Lazy;
 
 use std::{
-    cell::RefCell,
+    cell::{Cell, RefCell},
     cmp::Ordering,
     collections::{HashMap, HashSet},
     rc::Rc,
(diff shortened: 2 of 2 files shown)


61021:
Bump 🙂 — fixes :visible updates being silently lost for unmapped widgets. Happy to rebase if anything's drifted.

61021:
Bump. Still merges clean on master (48f5aa8) and `cargo check` passes on 1.81.

14 lines added and 6 removed in `widget_definitions.rs`, and it fixes #1451: a `:visible` update sent while a revealer is closed gets thrown away when the child finally maps. @elkowar if you would rather it were solved somewhere else, say so and I will rework it.

